Issue #5, opened 2023-01-26
opencost deploys to default namespace


mattray:
Despite creating the `opencost` namespace, the helm chart deploys to the `default` namespace.

PR #6, opened 2023-01-26
Deploy to the "opencost" namespace by default

mattray:
Match the behavior from https://www.opencost.io/docs/install

Fixes https://github.com/opencost/opencost-helm-chart/issues/5

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -22,7 +22,7 @@ opencost:
     name: opencost
 
   prometheus:
-    serviceName: my-prometheus
+    serviceName: my-prometheus-server
     namespaceName: prometheus
 
   ui:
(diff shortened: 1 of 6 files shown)


values.yaml
```diff
@@ -22,8 +22,8 @@ opencost:
     name: opencost
 
   prometheus:
-    serviceName: my-prometheus
-    namespaceName: opencost
+    serviceName: my-prometheus-server
+    namespaceName: prometheus
```
Pokom, line 26: Suggestion: Can we have these exposed as values so we can override the names?

Pokom (approved):
Just a minor suggestion on exposing the prometheus config as values. Otherwise :+1:

values.yaml
ts-mini: I did this in my latest PR https://github.com/opencost/opencost-helm-chart/pull/7

mattray: https://github.com/opencost/opencost-helm-chart/issues/8

PR #22, opened 2023-03-03
Extravolumes

alisaazimova:
Hello, 
We are running OpenCost on AKS and the ability to mount volumes would be extremely helpful. As of now, we are modifying the OpenCost chart, but it would be very appreciated if it could be added to an official chart. I believe it will help many users to solve their particular use cases.

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -1,4 +1,4 @@
-{{- include  "isPrometheusConfigValid" . }}
+{{- include  "isPrometheusConfigValid" . -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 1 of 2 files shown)


alisaazimova:
@toscott

toscott:
Hi there @alisaazimova. Thanks for the contribution!

I'm not a full maintainer for this project, but I am a reviewer and open to helping out.

- [ ] It looks like the current proposal doesn't supply anything in the values.yaml (or readme) to indicate how to supply the extra volumes
- [ ] It also looks like not specifying a value may cause some issues with installation as the parent keys aren't wrapped nicely for when nothing is provided

Can you provide some more information on what your use case is for mounting volumes for the wider audience?  I do like the idea and I could probably leverage this as well for supplying on prem pricing models at startup time instead of building them into a custom image.

toscott:
I guess one more note, the proposed values don't make it super clear that the extra volumes are intended for the opencost exporter container (vs the opencost-ui container). 

Perhaps something like this for the container volume attachment piece?
```
opencost:
  exporter:
    volumes: []
```

alisaazimova:
Yes, sorry, just updated values.yaml the way we are using it . Actually, we are taking an idea from kubecost helm chart [https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/templates/cost-analyzer-deployment-template.yaml](url) , we definitely can make reference to volumes more clear, to specify that it is mounted to opencost exporter container. Regarding our use cases, we mount some secrets from hashicorp vault enterprise ( for cloud integration)  plus some additional metrics-configs.

alisaazimova:
I added volumenMounts specifically to exporter

toscott:
Thanks for the updates @alisaazimova. I will try review & test this out this later today.

alisaazimova:
Thank you @toscott

mattray:
@toscott let me know when you're happy with this and we'll get it merged.

toscott (approved):
Looks good. Was able to install with and without volume mounts. 

Example config for a mounting a volume
```
opencost:
  exporter:
    extraVolumeMounts:
      - mountPath: /test
        name: test-volume
extraVolumes:
  - name: test-volume
    emptyDir:
      sizeLimit: 10Mi
```

mattray (approved):

PR #24, opened 2023-03-06
add pod annotation and make secret object optional

rabunkosar-dd:
- Adding ability to annotate pods
- Making secret creation optional based on the secret values
- Adding option to add bearer token for prometheus
- Updated values.yaml accordingly

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -20,6 +20,10 @@ spec:
       maxUnavailable: 1
     type: RollingUpdate
   template:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     metadata:
       labels:
         {{- include "opencost.selectorLabels" . | nindent 8 }}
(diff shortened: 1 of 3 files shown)


charts/opencost/templates/deployment.yaml
```diff
@@ -74,6 +78,13 @@ spec:
                   name: {{ include "opencost.fullname" . }}
                   key: DB_BASIC_AUTH_PW
             {{- end }}
+            {{- if .Values.opencost.prometheus.bearer_token }}
+            - name: DB_BEARER_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "opencost.fullname" . }}
+                  key: DB_BEARER_TOKEN
+            {{- end }}
```
Pokom, line 87: Only think I'd suggest here is to ensure only the `password` or the `bearer_token` can be set, and not both at once. Not a hard stop to getting this out, but would be nice to have.

rabunkosar-dd: good suggestion, I'll add that check

Pokom: Sounds good. Just re-request a review once you're ready and I can take another look.

Pokom (approved):
:+1: I'm assuming you've ran this @rabunkosar-dd?

rabunkosar-dd:
> 👍 I'm assuming you've ran this @rabunkosar-dd?

Thanks for pointing that out. I did a last minute run, looks like the pod annotation wasn't properly indented, which I didn't test properly. Fixed it now. Others are fine (username/password/bearer_token)

Pokom (approved):
:+1:

Pokom:
Going to revert because the publish stage failed. Root cause is we forgot to update the chart version.

PR #25, opened 2023-03-10
Bump version to 1.1.0 to be more strict on SemVer following.

mattray:
Added a feature, not a bugfix, so bumping minor version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 OpenCost and OpenCost UI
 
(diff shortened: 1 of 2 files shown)


toscott (approved):
Looks good.

Could probably use a helm-docs refresh of the chart readme as well at some point. I might make a new issue to go through and add descriptions for the non obvious items and generate them.

PR #27, opened 2023-03-10
feat: Install pinned version rather than latest tag

### LINKED ISSUE #23: Helm Chart image versioned with opencost released image
We should change the helm chart release to match the Opencost release.
At the moment in the values.yaml there is set "tag: latest". 
Unfortunately in this way, the helm chart cannot be versioned by controlling what app version image has been released within

Example of the requested change 
chart 1.0.3 has been set to 1.9.8 with the latest image 
    
    image:
      registry: quay.io
      repository: kubecost1/kubecost-cost-model
      tag: latest

to example:
chart 1.0.4 has been set to 1.100.0

    image:
      registry: quay.io
      repository: kubecost1/kubecost-cost-model
      tag: prod-1.100.2

mkilchhofer:
There was a mismatch between badges (AppVersion: 1.9.8) and version installed (latest tag).

While I fixed this, I documented every helm parameter via the required comments above each value. Let me know if I should split it into multiple PRs.

Resolves #23

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -1,90 +1,123 @@
 serviceAccount:
+  # -- Specifies whether a service account should be created
   create: true
+  # --  Annotations to add to the service account
   annotations: {}
   # eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/eksctl-opencost
+
+  # -- Whether pods running as this service account should have an API token automatically mounted
   automountServiceAccountToken: true
 
+# --  Annotations to add to the Deployment
 annotations: {}
-
+# --  Annotations to add to the OpenCost Pod
 podAnnotations: {}
 
 service:
+  # --  Annotations to add to the service
   annotations: {}
+  # --  Labels to add to the service account
   labels: {}
+  # --  Kubernetes Service type
   type: ClusterIP
 
 opencost:
   exporter:
-    # The GCP Pricing API requires a key. This is supplied just for evaluation.
-    # cloudProviderApiKey: 'asdfasdfasdf'
-    # Default cluster ID to use if cluster_id is not set in Prometheus metrics.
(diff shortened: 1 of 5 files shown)


charts/opencost/values.yaml
```diff
@@ -1,87 +1,119 @@
...
+      # -- Additional labels to add to the ServiceMonitor
       additionalLabels: {}
-      ## The label to use to retrieve the job name from.
-      ## jobLabel: "app.kubernetes.io/name"
-      namespace: ''
-      namespaceSelector: {}
-      ## Default: scrape .Release.Namespace only
-      ## To scrape all, use the following:
-      ## namespaceSelector:
-      ##   any: true
```
mkilchhofer, line 48: Removed `namespaceSelector` from values, as template does not honor this.

charts/opencost/values.yaml
```diff
@@ -1,87 +1,119 @@
...
+      # -- Specify if the ServiceMonitor will be deployed into a different namespace (blank deploys into same namespace as chart)
+      namespace: ""
+      # -- Interval at which metrics should be scraped
       scrapeInterval: 30s
-      # honorLabels: true
-      targetLabels: []
```
mkilchhofer, line 51: Removed `targetLabels` from values, as template does not honor this.

mkilchhofer:
> I've got a suggestion to add to your work. Probably we should remove any "default" env variable and let the people/engineer decide with valuesinline what set like example this one https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/templates/deployment.yaml#L65 . What do you think?

Yeah this can be discussed in further PRs. I have several additional changes I'd like to propose but I will do that in a different PR. ;) In my day-to-day job we have a Code of conduct where we should only file the changes that are required to a specific feature. :-)

toscott (approved):
Looks good @mkilchhofer  - Thanks for the contribution!

PR #28, opened 2023-03-10
fix,chore: Plain text of API key

hakuno:
I use GitOps tool like FluxCD or ArgoCD. So, the code gets versioned.

So I needed to put an API key of GCP... but the `opencost.exporter.cloudProviderApiKey` is vulnerable. My security scanner argues about it.

Usually we could use a Kubernetes Secret that can be read from an Env. Like this

```
opencost:
  exporter:
    defaultClusterId: 'default-cluster'
    env:
    - name: CLOUD_PROVIDER_API_KEY
      valueFrom:
        secretKeyRef:
          name: some-safety-secret
          key: key
```

It's fine. It's standard. So I'm sure the OpenCost Helm Chart can support it too.

Ah! I tested it with no problem, you can do so.

```
helm template --dry-run .
```

Please, I need this fix/feature. Can you help me out?

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 appVersion: 1.9.8
 name: opencost
 description: OpenCost and OpenCost UI
-version: 1.1.0
+version: 1.1.1
 maintainers:
   - name: mattray
     url: https://mattray.dev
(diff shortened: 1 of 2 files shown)


mattray:
I merged this forward over here: https://github.com/opencost/opencost-helm-chart/pull/36
Once I get a review I'll merge that and close this.

PR #31, opened 2023-03-13
fix the location of annotations for pods

rabunkosar-dd:
Fixing the incorrect location for the pod annotation in https://github.com/opencost/opencost-helm-chart/pull/24 and also bumped the version to avoid the version issues.

@Pokom I noticed that issue while deploying, so apologies for the back and forth, can you take a look please?

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -23,9 +23,9 @@ spec:
     metadata:
       labels:
         {{- include "opencost.selectorLabels" . | nindent 8 }}
-      {{- with .Values.podAnnotations }}
+      {{- if .Values.podAnnotations }}
       annotations:
-        {{- toYaml . | nindent 8 }}
+      {{- toYaml .Values.podAnnotations | nindent 8 }}
       {{- end }}
     spec:
     {{- if .Values.imagePullSecrets }}
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -20,13 +20,13 @@ spec:
       maxUnavailable: 1
     type: RollingUpdate
   template:
-    {{- with .Values.podAnnotations }}
-    annotations:
-      {{- toYaml . | nindent 8 }}
-    {{- end }}
     metadata:
       labels:
         {{- include "opencost.selectorLabels" . | nindent 8 }}
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
     {{- if .Values.imagePullSecrets }}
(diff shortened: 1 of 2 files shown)


charts/opencost/templates/deployment.yaml
```diff
@@ -20,13 +20,13 @@ spec:
...
-    annotations:
-      {{- toYaml . | nindent 8 }}
-    {{- end }}
     metadata:
       labels:
         {{- include "opencost.selectorLabels" . | nindent 8 }}
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
Pokom, line 29: Ah I see, the annotations was not in the correct spot. Hard to spot that in a review, but good catch.

Pokom (approved):
:+1: From me. Again this is a bit harder for me to test out manually, so let me know once this has been deployed out.

charts/opencost/templates/deployment.yaml
```diff
@@ -20,13 +20,13 @@ spec:
...
-    {{- with .Values.podAnnotations }}
-    annotations:
-      {{- toYaml . | nindent 8 }}
-    {{- end }}
     metadata:
       labels:
         {{- include "opencost.selectorLabels" . | nindent 8 }}
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
```
Pokom, line 28: Is it okay to not be explicit here with what we're writing out? Most other places in the codebase we're being explicit. IE, I believe it is a bit clearer if we say:

```
annotations:
{{- toYaml .Values.podAnnotations | nindent 8 }}
```

rabunkosar-dd: I've made the change, here is the test:

value:
```
podAnnotations:
  "iam.amazonaws.com/role": arn:aws:iam::12345:role/a
```

output:
```
    metadata:
      labels:
        app.kubernetes.io/name: opencost
        app.kubernetes.io/instance: opencost
      annotations:
        iam.amazonaws.com/role: arn:aws:iam::12345:role/a
```

the default case:

```
    metadata:
      labels:
        app.kubernetes.io/name: opencost
        app.kubernetes.io/instance: opencost
```

Hopefully the YAML is properly formatted and the objects are now in the right place 🤞

Pokom: Double checking against other deployments I've got with annotations and that looks like it's in the correct spot! Thanks for making the change

Pokom (approved):
:+1: Tested personally with 

```
helm template --set opencost.prometheus.internal.enabled=false --set opencost.prometheus.external.enabled=true --set opencost.metrics.serviceMonitor.enabled=true --set podAnnotations="test.this=foo" .
```

And noted that the annotation was created in the correct location with `test.this=foo`

PR #38, opened 2023-03-15
feat: Add ability to define securityContexts and podSecurityContext

mkilchhofer:
This PR adds the ability to define the securityContext on both container- and pod-level.

/cc: @toscott @mattray @Pokom

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -13,6 +13,10 @@ annotations: {}
 # --  Annotations to add to the OpenCost Pod
 podAnnotations: {}
 
+# -- Holds pod-level security attributes and common container settings
+podSecurityContext: {}
+  # fsGroup: 2000
+
 service:
   # --  Annotations to add to the service
   annotations: {}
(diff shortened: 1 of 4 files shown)


mattray:
This looks good to me, my understanding is we're just putting in the support for the values to be user-supplied and aren't providing any defaults, which is fine IMO. If I get confirmation of my understanding we'll merge.

charts/opencost/values.yaml
```diff
@@ -111,6 +124,14 @@ opencost:
       limits:
         cpu: '999m'
         memory: '1Gi'
+    # -- The security options the container should be run with
+    securityContext: {}
+      # capabilities:
+      #   drop:
+      #   - ALL
+      # readOnlyRootFilesystem: true
```
toscott, line 132: Just as a note to anyone leveraging this later, the UI container crashes if readOnlyRootFilesystem is true as it needs to write to tmp space.

[emerg] 1#1: mkdir() "/tmp/proxy_temp" failed (30: Read-only file system)

toscott (approved):
lgtm. Thanks for another contribution @mkilchhofer

toscott:
> This looks good to me, my understanding is we're just putting in the support for the values to be user-supplied and aren't providing any defaults, which is fine IMO. If I get confirmation of my understanding we'll merge.

Agreed. This is providing additional functionality, but does not enable any of the pod security constraints by default.

PR #35, opened 2023-03-14
chore: Adapt workflow to common helm testing

mkilchhofer:
I ([argo-helm](https://github.com/argoproj/argo-helm) maintainer) would like to share our workflow for CI testing.

Resolves #30

Tested it on my fork repo:
grafik

The code is mostly based on the howto/README of the [helm/chart-testing-action](https://github.com/helm/chart-testing-action).
The only thing we (at argo-helm) recently adapted is to remove the deprecated `echo "::set-output name=changed::true"` with `echo "changed=true" >> $GITHUB_STATE`.

Deprecation and howto fix is documented in the GitHub blog post: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples

PR #39, opened 2023-03-15
fix: Last PR (#35) contained a bug, replace GITHUB_STATE -> GITHUB_OUTPUT

mkilchhofer:
Hi again

My last PR (#35) contained a bug which was also present in my main repository over there: [argoproj/argo-helm](https://github.com/argoproj/argo-helm).
So this one here is a "backport" of our fix:
- https://github.com/argoproj/argo-helm/pull/1904

/cc: @mattray @Pokom @toscott


toscott (approved):
Reviewed docs here on github state vs github output. 

The way its being accessed in the Create kind cluster and run chart testing aligns with $GITHUB_OUTPUT environment variable.

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

Thanks for the fix @mkilchhofer

Issue #30, opened 2023-03-11
Add CI check to ensure PR's bump versions


Pokom:
Currently it is possible to merge changes into `main` that do not bump the `appVersion` in the `Charts.yaml` file. There should be some safeguards in place to prevent this to ensure `main` can always be published.

mkilchhofer:
@mattray @sntxrr you can assign that to me 🙋‍♂️

mkilchhofer:
@Pokom Thx for assigning it to me 👍 PR is ready. Shared the CI workflow of [argo-helm](https://github.com/argoproj/argo-helm) where I invest my main contributor/maintainer time :-)

Pokom:
Awesome thank you @mkilchhofer!!! I can take a look when I have a moment

PR #41, opened 2023-03-16
feat: add Ingress for opencost UI

### LINKED ISSUE #32: Add Ingress for UI
Just a basic feature.

boniek83:
Fixes #32

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -136,15 +136,25 @@ opencost:
     # -- A list of volume mounts to be added to the pod
     extraVolumeMounts: []
 
-    # -- Ingress for OpenCost UI
     ingress:
+      # -- Ingress for OpenCost UI
       enabled: false
+      # -- Ingress controller which implements the resource
       ingressClassName: ""
+      # -- Annotations for Ingress resource
+      annotations: {}
+        # kubernetes.io/tls-acme: "true"
+      # -- A list of host rules used to configure the Ingress
+      # @default -- See [values.yaml](values.yaml)
       hosts:
         - host: example.local
           paths:
             - /
- 
+      # -- Ingress TLS configuration
+      tls: []
+        #  - secretName: chart-example-tls
+        #    hosts:
+        #      - chart-example.local
   # -- Toleration labels for pod assignment
   tolerations: []
   # -- Node labels for pod assignment
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -136,6 +136,15 @@ opencost:
     # -- A list of volume mounts to be added to the pod
     extraVolumeMounts: []
 
+    # -- Ingress for OpenCost UI
+    ingress:
+      enabled: false
+      ingressClassName: ""
+      hosts:
+        - host: example.local
+          paths:
+            - /
+ 
   # -- Toleration labels for pod assignment
   tolerations: []
   # -- Node labels for pod assignment
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - kubecost
   - opencost
   - monitoring
-version: 1.7.0
+version: 1.8.0
 maintainers:
   - name: mattray
     url: https://mattray.dev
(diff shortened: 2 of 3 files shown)


mkilchhofer (changes requested):
Hi @boniek83 

I am only a contributor here and not a maintainer so my review is not that relevant ;-)

But I'd add an Ingress template which is almost identical with the `helm create mychart` template/skeleton.
The reason behind this is that OpenCost is per definition supported from Kubernetes versions >1.8:

> **What Kubernetes versions does OpenCost support?**
> OpenCost supports a Kubernetes environment above 1.8.
>
> Source: https://www.opencost.io/docs/FAQ#what-kubernetes-versions-does-opencost-support

Therefore the skeleton would be perfect backward compatible:

```yaml
{{- if .Values.ingress.enabled -}}
{{- $fullName := include "foobar.fullname" . -}}
{{- $svcPort := .Values.service.port -}}
{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
  {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
  {{- end }}
{{- end }}
{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
apiVersion: networking.k8s.io/v1
{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
apiVersion: networking.k8s.io/v1beta1
{{- else -}}
apiVersion: extensions/v1beta1
{{- end }}
kind: Ingress
metadata:
# ...
```

mattray:
@boniek83 do you want to merge in @mkilchhofer's suggestion? We'll merge afterwards

mkilchhofer:
> @boniek83 do you want to merge in @mkilchhofer's suggestion? We'll merge afterwards

@mattray where does this statement comes from. Do you still support kubernetes 1.8 as of 2023? ;) 
As the chart does not follow this today.

In kubernetes 1.8, the following APIs are not available:
- Deployment: `apps/v1` is not available (needs to be `apps/v1beta1`)
- ClusterRoles / ClusterRoleBindings: `rbac.authorization.k8s.io/v1` not available (needs to be `rbac.authorization.k8s.io/v1beta1`)
- ...

boniek83:
@mattray I can do that, but you don't support kubernetes 1.8 anywhere else in this helm chart, so I wonder if there is a point to that. Things can always be changed later if issue of compatibility pops up, which is not likely IMHO. Additionally is it worth it to support such old versions? Helm chart will get increasingly complicated as it gains more features and as kubernetes versions keep getting bigger.

mattray (approved):

mattray:
Sorry, I blanked on that thinking of another recent breaking change and my knee-jerk response was "backwards compatibility good" ;)

boniek83:
I will fix conflict tomorrow.

boniek83:
@mattray  Maintainers will have to bump helm chart version or should I do it?

mattray:
@boniek83 you can do it or we (the maintainers) will do it manually after merging. It would make sense for us to automate that too.

mkilchhofer:
> @boniek83 you can do it or we (the maintainers) will do it manually after merging. It would make sense for us to automate that too.

🤔 I automated it in:
- #35
- #30

Contributors should bump it IMHO. This is the standard workflow for chart contributions in the helm community.

charts/opencost/values.yaml
```diff
@@ -136,6 +136,15 @@ opencost:
...
 
+    # -- Ingress for OpenCost UI
+    ingress:
+      enabled: false
+      ingressClassName: ""
+      hosts:
+        - host: example.local
+          paths:
+            - /
+
```
mkilchhofer, line 147: Please document all possible values separately:

```suggestion
    ingress:
      # -- Ingress for OpenCost UI
      enabled: false
      # -- Ingress controller which implements the resource
      ingressClassName: ""
      # -- Annotations for Ingress resource
      annotations: {}
        # kubernetes.io/tls-acme: "true"
      # -- A list of host rules used to configure the Ingress
      # @default -- See [values.yaml](values.yaml)
      hosts:
        - host: example.local
          paths:
            - /
      # -- Ingress TLS configuration
      tls: []
        #  - secretName: chart-example-tls
        #    hosts:
        #      - chart-example.local
```

mkilchhofer (changes requested):
After implementing my proposed changes, please run helm-docs again.

PR #47, opened 2023-03-27
add secret annotation to enable reading from Vault

Qixuanw:


toscott (approved):
Thanks for the contribution @Qixuanw!

PR #51, opened 2023-03-28
fix: invalid helm template for ingress annotations

piotrjanik:
There is a bug in ingress.yaml helm template for annotations.

To reproduce bug uncomment ingress configuration in the values.yaml file(or just use --set).

Full path should not be used within "with" block.


toscott (approved):
Verified annotations are working as expected with this change.

toscott:
Thanks for the contribution @piotrjanik!

PR #54, opened 2023-04-03
Sets honorLabels default value to true. Fixes #53

### LINKED ISSUE #53: Non-working default in ServiceMonitor value
When `honorLabels` is set to `false`, OpenCost metrics are being ingested with the `namespace` label set to the opencost namespace. This causes most of the data to be set to `unallocated`, as the ingested namespace labels do not match the pods.

edmeister:


mattray:
I see in [Kubecost this defaults to "true"](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/values.yaml#L531), so I think I'm good with this. Could you update the README.md to document the change in the default?

edmeister:
> 

Done

charts/opencost/Chart.yaml
```diff
@@ -9,7 +9,7 @@ keywords:
   - kubecost
   - opencost
   - monitoring
-version: 1.10.0
+version: 1.11.0
```
toscott, line 12: @edmeister - This version still needs bumped to 1.12.0 for CI checks to pass. Doesn't look like I can edit your branch or I would. Sorry!

toscott:
Will pull into a bulk merge and update version.

toscott:
Merging here:
https://github.com/opencost/opencost-helm-chart/pull/60

PR #59, opened 2023-04-12
chore: liveness and readiness probes

piotrjanik:
This adds readiness and liveness probes for both exporter and ui on `/healthz` path.


toscott:
Merging here:
https://github.com/opencost/opencost-helm-chart/pull/60

PR #65, opened 2023-05-08
Update PVC claim name

lupa95:
Hi, this PR will fix the broken pvc setup in the Helm chart.

If you set `persistence.enabled: true` in the values.yaml, the opencost container will be pending for the PVC. The helm-chart creates a PVC with a different name and it is not configurable. This should align the two names and fix the issue.


toscott (approved):
Thanks for the contribution @lupa95! This will be available in version 1.14.3.

PR #68, opened 2023-05-10
fix: increase chart version with Alto's prometheus endpoint removal

eze-kiel:
This fixes the broken test from commit 6ed2eb4


toscott (approved):
lgtm

Issue #70, opened 2023-05-17
The secret is not created even if I put in the aws settings.


minoru7304:
You can create a secret in the "aws" setting by modifying as follows
`{{- if or .Values.opencost.prometheus.username .Values.opencost.prometheus.password .Values.opencost.prometheus.bearer_token .Values.opencost.exporter.aws.access_key_id -}}`

PR #72, opened 2023-05-18
fix: Secret is created by adding aws settings

minoru7304:
The following will be resolved
https://github.com/opencost/opencost-helm-chart/issues/70


toscott (approved):
Thanks for the contribution @minoru7304! This will be available in the chart version: 1.14.2

PR #74, opened 2023-05-25
feat: Add network-cost daemonset to chart

Reasonably:
# Problem
- Opencost helm chart doesn't have network cost daemonset.
- Opencost have feature calculating network cost, It cannot use this feature without network cost daemonset

# Description
- Add network-costs daemonset.
- It derived almost from https://github.com/kubecost/cost-analyzer-helm-chart.


mattray:
The network-cost daemon isn't part of OpenCost, it's part of the Kubecost commercial product. Given that it's not open source and not part of OpenCost, I'm closing this PR.

PR #79, opened 2023-06-09
Remove redundant namespace

dnskr:
The PR removes redundant `namespace` from the chart templates.

Helm creates Kubernetes resources (Deployment, ConfigMap, Service, Ingress etc.) in the same namespace where the chart installed, i.e. `.Release.Namespace` is used by default. So there is no reason to set namespace explicitly and it can be removed to keep code cleaner.

### DIFF
--- a/charts/opencost/templates/clusterrole.yaml
+++ b/charts/opencost/templates/clusterrole.yaml
@@ -6,7 +6,6 @@ metadata:
   labels:
     {{- include "opencost.labels" . | nindent 4 }}
   name: {{ include "opencost.fullname" . }}
-  namespace: {{ .Release.Namespace }}
 rules:
   - apiGroups: [""]
     resources:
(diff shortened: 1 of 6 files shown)


dnskr:
@mattray @toscott Could you please have a look at the PR? Sorry for pinging you directly.

mattray:
If you leave off the `--namespace opencost` using `helm install opencost opencost/opencost` you'll create the Deployment, Service, ClusterRole, etc. in the `default` namespace. We hardcoded `opencost` over in the [default manifest](https://github.com/opencost/opencost/blob/develop/kubernetes/opencost.yaml) for deploying OpenCost, should we expect Helm users to provide their namespace? I can go either way.

mattray:
I've tested the patch and it seems like [the consensus](https://github.com/helm/helm/issues/5465) is to provide `--namespace` from the CLI, so I'm merging.

mattray (approved):

Issue #82, opened 2023-06-20
ui authentication


TheRealNoob:
it would be great to be able to have authentication of some kind on the webui so that you can safely expose it to the internet.  i was thinking about an oauth2-proxy sidecar.

mattray:
That sounds great, we'd certainly appreciate it.

mattray:
The PR needs more testers.

PR #84, opened 2023-06-23
add support for oauth-proxy sidecar

TheRealNoob:
Add the ability to add sidecars with the specific use-case in mind of oauth-proxy for the UI.

closes https://github.com/opencost/opencost-helm-chart/issues/82


TheRealNoob:
one thing I think worth noting is that I nested `extraContainers` at a different level than `extraVolumes`.  This is obviously not ideal, but I did it this way because I believe extraVolumes is located at the wrong level.  Let me explain.

If I could be so forward, what I see implemented now a mixed presumption that this helm chart will only ever have one Pod and ServiceMonitor (because a SM can only point to a single pod).  Some of the config says that statement is true and some says it's false.  On the root of the values file are several items:
* podAnnotations
* podLabels
* podSecurityContext
* extraVolumes
* etc
All of these objects can only be tied to one podSpec.  If you add a second pod down the road, now these objects being located at the root is confusing.  I think you lock yourself into a corner by having them there - I think the correct place is under the `opencost` object.  An example of a second pod that you might add down the road could be subcharting Prometheus, in which case that would be controlled under the root `prometheus` object.  I'm tempted to say that everything written now should be moved under `opencost`.  Even things like ServiceAccount, which isn't technically 1:1, but if you're going to create another pod, it probably won't need the same k8s permissions as opencost, so having a different serviceAccount for it makes sense IMO.

This is all just my internal thoughts as I was drafting it.  If you want me to move things around let me know, I won't object.

mattray:
I don't have a means to test this, but if someone could validate this works for them we'll get it merged.

debMan:
@TheRealNoob can you rebase it to the master?

charts/opencost/templates/servicemonitor.yaml
```diff
@@ -26,6 +26,9 @@ spec:
     {{- if .Values.opencost.metrics.serviceMonitor.metricRelabelings }}
       metricRelabelings: {{ toYaml .Values.opencost.metrics.serviceMonitor.metricRelabelings | nindent 8 }}
     {{- end }}
+  {{- with .Values.opencost.metrics.serviceMonitor.extraEndpoints }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
debMan, line 31: Not related to this PR.

debMan (approved):
Works fine.

mattray (approved):

PR #86, opened 2023-06-26
Missing chart url in docs

davhdavh:


toscott:
Hi @davhdavh. Thanks for trying to improve our documentation!

The documentation for adding the opencost-helm-chart repository lives at the root of this project. In the future, we may have multiple helm charts produced by this repository and I believe we would like to leave instructions on how to add the repo in one spot. 

Going to close this one.

davhdavh:
FYI, it took me 1h to find the answer to this, because it isnt just easily googleable, and if you dont document it in the readme, the docs on artifacthub.io will just leave anyone to wonder where the chart rep is...
It is also very much against the standard to charts available on artifacthub.io that all document what the chart url is

PR #92, opened 2023-07-06
Added name field in service account

rohitraut3366:
Added service account name field in `values.yaml`

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -9,7 +9,9 @@ serviceAccount:
   # --  Annotations to add to the service account
   annotations: {}
   # eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/eksctl-opencost
-
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
   # -- Whether pods running as this service account should have an API token automatically mounted
   automountServiceAccountToken: true
 
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - kubecost
   - opencost
   - monitoring
-version: 1.16.2
+version: 1.17.2
 maintainers:
   - name: mattray
     url: https://mattray.dev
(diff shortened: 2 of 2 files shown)


mattray:
Did you mean to patch https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/templates/serviceaccount.yaml#L5? You'll need to bump the version again, it's currently 1.17.1

rohitraut3366:
This change will make things clear at the values.yaml file level. so users don't have to go in and check how the service account name is being generated.

charts/opencost/values.yaml
```diff
@@ -9,7 +9,9 @@ serviceAccount:
   # --  Annotations to add to the service account
   annotations: {}
   # eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/eksctl-opencost
-
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
```
mattray, line 14: I'm not sure serviceAccount.name would get applied, could you verify with
https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/templates/serviceaccount.yaml#L5

rohitraut3366: Yes, the service name is getting applied correctly with the name field. here is the ref to code https://github.com/rohitraut3366/opencost-helm-chart/blob/rohit/added-name-field-in-service-account/charts/opencost/templates/_helpers.tpl#L72

dnskr: The following command tests that `.Values.serviceAccount.name` get applied correctly:
```
helm template my-test charts/opencost --set serviceAccount.name=custom-name
```
Output
```
---
# Source: opencost/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: custom-name
  labels:
    ...
automountServiceAccountToken: true
---
# Source: opencost/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-test-opencost
  labels:
    ...
spec:
  ...
  template:
    metadata:
      labels:
        ...
    spec:
      serviceAccountName: custom-name
      containers:
```

mattray (approved):

PR #100, opened 2023-08-08
Template Prometheus values

dnskr:
The PR templates external and internal Prometheus values.
The change is needed to be able to use dynamic values.

Before the change (output truncated)
```yaml
helm template test charts/opencost --set opencost.prometheus.internal.enabled=false --set opencost.prometheus.external.enabled=true --set opencost.prometheus.external.url="\{\{ \"test\" | upper \}\}" --show-only templates/deployment.yaml
---
# Source: opencost/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-opencost
spec:
  template:
    spec:
      containers:
        - name: test-opencost
          env:
            - name: PROMETHEUS_SERVER_ENDPOINT
              value: "{{ \"test\" | upper }}"
```
After the change (output truncated)
```yaml
helm template test charts/opencost --set opencost.prometheus.internal.enabled=false --set opencost.prometheus.external.enabled=true --set opencost.prometheus.external.url="\{\{ \"test\" | upper \}\}" --show-only templates/deployment.yaml
---
# Source: opencost/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-opencost
spec:
  template:
    spec:
      containers:
        - name: test-opencost
          env:
            - name: PROMETHEUS_SERVER_ENDPOINT
              value: "TEST"
```


dnskr:
Is there any reason or use case to have both `external` and `internal` sections?
I think that current implementation can be simplified, i.e. the following
```yaml
  prometheus:
    ...
    external:
      enabled: false
      url: "https://prometheus.example.com/prometheus"
    internal:
      enabled: true
      serviceName: my-prometheus
      namespaceName: opencost
      port: 9090
```
can be replaced with
```yaml
  prometheus:
    ...
    url: "http://prometheus-server.opencost.svc:9090"
```

I can submit a new PR if the change make sense.

dnskr:
@mattray @toscott Could you please review the PR and also have a look into my comment/question?

mattray (approved):

PR #107, opened 2023-08-27
Add customization option for metrics config and cluster info

tamalsaha:


NItishSh (approved):

mattray:
There's a lot going on in this PR. Could you please open an issue (or 2) explaining what your goals are with this?

tamalsaha:
@mattray PTAL. It nows creates a configmap to mount metrics.json and cluster-info.json

mattray:
I think I like the approach of https://github.com/opencost/opencost-helm-chart/pull/109 even though it came after this patch. I'll probably merge that one and close this one after they make the requested changes (or this one if you bring in their changes).

charts/opencost/templates/deployment.yaml
```diff
@@ -1,3 +1,7 @@
+{{- if not (hasKey .Values.opencost.exporter.extraEnv "KUBECOST_NAMESPACE") }}
+{{- $_ := set .Values.opencost.exporter.extraEnv "KUBECOST_NAMESPACE" .Release.Namespace }}
+{{- end }}
+
```
mattray, line 4: we're not using Kubecost, should this be "OPENCOST_NAMESPACE"?

PR #110, opened 2023-08-28
Explicitly set the namespace field in the yaml

tamalsaha:
Why?
https://www.reddit.com/r/kubernetes/comments/uglbn3/should_i_define_the_namespace_metadata_info_on_my/

If `helm template | kubectl -f -` is used, this pr will ensure the desired namespace is used.

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -3,6 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "opencost.fullname" . }}
+  namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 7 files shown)


mattray:
We had a previous PR to remove this: https://github.com/opencost/opencost-helm-chart/pull/79

The consensus was that we should provide the namespace from the CLI, it made automating easier.

tamalsaha:
> The consensus was that we should provide the namespace from the CLI, it made automating easier.

Can you please show me an example of such command? I am trying to follow `helm template | kubectl apply -f -` format.

mattray:
You can use `--namespace opencost` 
https://helm.sh/docs/helm/helm_template/

Issue #73, opened 2023-05-22
kubecost-cost-model container not starting up in Azure when deployed with OpenCost Helm chart.


ckarwicki:
Hi,

It appears `kubecost-cost-model` container is not starting up when deployed in Azure AKS with Opencost Helm chart. There are no errors in logs, even changing `LOG_LEVEL` to `debug` didn't have any effect (probably because container doesn't even start). We suspect there is something with the OpenCost Helm chart incorrectly configuring deployment of  `kubecost-cost-model`.  Tried many things but no difference, same log Info and container not started:
1. Helm chart 1.14.0, regular `quay.io/kubecost1/kubecost-cost-model:prod-1.103.1` and with `gcr.io/kubecost1/opencost:kc-eu-2023`
2. Helm chart 1.14.3,  regular `quay.io/kubecost1/kubecost-cost-model:prod-1.103.1` and with `gcr.io/kubecost1/opencost:kc-eu-2023`
3. With Service Principal secret and without.

Details:
Azure AKS v1.25.6
Deployed with ArgoCD v2.6.7+5bcd846
```
project: ska
source:
  repoURL: 'https://opencost.github.io/opencost-helm-chart'
  targetRevision: 1.14.3
  helm:
    values: |
      extraVolumes:
      - name: service-key-secret
        secret:
          secretName: azure-service-key
      opencost:
        exporter:
          extraEnv:
            LOG_LEVEL: debug
          extraVolumeMounts:
          - mountPath: /var/secrets
            name: service-key-secret
        prometheus:
          internal:
            enabled: true
            namespaceName: monitoring
            port: 9090
            serviceName: prometheus-kube-prometheus-prometheus
        ui:
          ingress:
            annotations:
              cert-manager.io/cluster-issuer: letsencrypt-prod
              ingress.kubernetes.io/ssl-redirect: "true"
              kubernetes.io/tls-acme: "true"
              nginx.ingress.kubernetes.io/modsecurity-snippet: |
                SecRuleEngine On
            ingressClassName: nginx
    version: v3
  chart: opencost
destination:
  server: 'https://kubernetes.default.svc'
  namespace: kubecost
syncPolicy:
  automated:
    prune: true
    selfHeal: true
  syncOptions:
    - CreateNamespace=true
    - ApplyOutOfSyncOnly=true
  retry:
    limit: 5
```
Logs, only this is in logs no matter what is changed:
```
27
[...]

PR #112, opened 2023-08-30
increased readiness and liveness probe delay to 120

TheisFerre:
Increased default values `initialDelaySeconds` for `livenessProbe` and `readinessProbe`. Related to #73

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - kubecost
   - opencost
   - monitoring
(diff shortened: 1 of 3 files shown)


mattray (approved):

mattray (approved):

dwbrown2:
I'm surprised this would be needed given that Kubecost never had to increase past 30 seconds. Are we potentially masking some other issue with this change?

mattray:
My assumption was we were deploying to installations with less resources, but we should keep an eye out.

Issue #97, opened 2023-08-02
No data for "App cost by pod" panel in grafana example dashboard


thiDucTran:
I imported the dashboard @ https://github.com/opencost/opencost-helm-chart/blob/main/examples/dashboard/kube-prometheus-stack-opencost-dashboard.json to Grafana (kube-prometheus stack)

Seems metrics related to pods or containers are not showing

![image](https://github.com/opencost/opencost-helm-chart/assets/37130368/204b2903-ae7e-4e79-b151-ec4137582b91)

I tried editing some existing panel to use `kube_node_labels{job="opencost"}` instead of just `kube_node_labels` as commented in https://github.com/opencost/opencost-helm-chart/pull/9#issuecomment-1469921616 but does not seem to fix anything

* AKS: v1.25
* quay.io/kubecost1/kubecost-cost-model:prod-1.105.1
* kube-prometheus v0.12.0

thiDucTran:
Am I doing something wrong because the metrics, `container_memory_allocation_bytes` and `container_cpu_allocation`, do not have the `exported_container` label...they do the a `container` label....so if I edit to something like below...I do then see data in the panel

```
  (sum(sum(container_memory_allocation_bytes{exported_namespace=~"$namespace", container=~"$app.*"}) by (container,exported_instance) * on(exported_instance) group_left() (
				node_ram_hourly_cost{} / 1024 / 1024 / 1024
				+ on(node,instance_type) group_left()
					label_replace
					(
						kube_node_labels{job="opencost"}, "instance_type", "$1", "label_node_kubernetes_io_instance_type", "(.*)"
					) * 0
			)
  + 
  sum(container_cpu_allocation{exported_namespace=~"$namespace", container=~"$app.*"}) by (container,exported_instance) * on(exported_instance) group_left() (
	  			node_cpu_hourly_cost{} + on(node,instance_type) group_left()
		  			label_replace
		  			(
		  				kube_node_labels{job="opencost"}, "instance_type", "$1", "label_node_kubernetes_io_instance_type", "(.*)"
		  			) * 0
		  	))) * 24 * 30
```

mattray:
If someone with more Grafana experience wants to weigh in, I'm sure we'd appreciate a PR to fix this.

salvareztw:
Checking the prometheus metrics it seems to just have a wrong label applied as you suspected @thiDucTran so the Prom query is failing.  Other panels in the example dashboard don't have this problem, I can submit a PR on it

PR #114, opened 2023-09-05
#97 Remove exported_ prefix from metric labels in failing panels

ghost:
Removed the "exported_" prefix on some metric labels in panels throughout the example dashboard as it was causing the Prometheus queries to fail and no data to be rendered.


spalvarez:
I submitted this PR under an account I meant to delete.  let me know if I should re-submit from this GH account that ownership of the fork was transferred to

toscott:
This one was bundle merged in #117

PR #116, opened 2023-09-19
Add scrapeTimeout to values.yaml with 10s default

liorfranko:


toscott:
This one was bundle merged in #117

PR #123, opened 2023-10-03
Store and reference cloud API key in secret

namm2:
The `cloudProviderApiKey` value should be stored in a secret and referenced from this secret than exposing it. IT Sec people would thank for this.


mattray (approved):
Tested locally, works great. Thanks!

PR #135, opened 2023-10-25
Allow overriding the pricing provider type

toscott:
What
Allow overriding the pricing provider type in the custom pricing configuration file. 

Why
This allows us to provide the costModels for any provider directly through the configmap in the helm chart. For example, the aws provider and spot feed information could be configured with a values override that looks like this:

```
opencost:
  customPricing:
    enabled: true
    provider: aws    
    costModel:
      awsSpotDataRegion: us-east-2
      awsSpotDataBucket: my-spot-bucket
      awsSpotDataPrefix: spot_pricing_prefix
```

My hope is this simplifies Cost Model configuration for all providers by providing a straightforward path to configure it directly through the helm chart.

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - kubecost
   - opencost
   - monitoring
-version: 1.22.2
+version: 1.22.3
 maintainers:
   - name: mattray
     url: https://mattray.dev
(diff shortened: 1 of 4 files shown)


mattray (approved):
This seems like a pretty good approach going forward.

mattray:
You'll need to rebase and we should circle back on the documentation to add a section on custom pricing (beyond just on-prem).

PR #138, opened 2023-10-27
Proposal: add unit tests

sntxrr:
- as part of my jobs internal rigor around helm, we require unit testing
- thought it was worth proposing adding unit tests!


toscott (approved):
I like it. Something we could get added to the build pipeline as well and cover all of the functions in _helpers as well.

PR #145, opened 2023-11-03
Add support for opencost.cloudIntegrationSecret

mattray:
We can't expect it to always be passed from the CLI


ThiagoFelippi:
Some updates about the approval? We need the **cloudIntegrationSecret** to stop passing the secret via values.
We use ArgoCD, and pass the API Key hardcoded.

mattray:
@ThiagoFelippi if you want to test the patch and submit a review, that will work.

lmello (approved):
Approved.

PR #152, opened 2023-11-20
add in fullImageName to support development envs

saweber:
In order to use [tilt](tilt.dev), OpenCost's helm chart needs to support a value that is the 'fullImageName' of a docker image, to increase flexibility for development environments - particularly the case of images going to a locally run cluster (like kind or minikube) and do not need to be pushed to docker.hub or another image repository.

This is also designed to align with an identical feature in the Kubecost helm charts.


AjayTripathy:
@mattray can you help triage?

mattray (approved):

PR #151, opened 2023-11-19
Add initial support for network policies

DonaldKellett:
## What does this PR change?

This PR introduces initial support for network policies

## Does this PR relate to any other PRs?

Inspired by #56 

## How will this PR impact users?

Adds the ability to enable network policies simply by setting the field `networkPolicies.enabled` to `true`

## Does this PR address any GitHub or Zendesk issues?

Closes #21

## How was this PR tested?

- kind version: `v0.20.0`
- K8s version \(both client and server\): `v1.27.3`
- Helm version: `v3.13.2`

### Internal Prometheus

- Repository: `https://prometheus-community.github.io/helm-charts`
- Chart: `prometheus`
- Release name: `prometheus`
- Namespace: `prometheus-system`

### OpenCost configuration

1. Using the defaults in `charts/opencost/values.yaml`
1. With `networkPolicies.enabled` set to `true`

## Does this PR require changes to documentation?

Included within the PR, please review

## Have you labeled this PR and its corresponding Issue as "next release" if it should be part of the next OpenCost release? If not, why not?

Not applicable to this Helm chart repository, I believe?

PR #155, opened 2023-11-30
Rebased #151 with version bump

mattray:
Tested and verified https://github.com/opencost/opencost-helm-chart/pull/151, this is the version bump


charts/opencost/values.yaml
```diff
@@ -241,11 +256,11 @@ opencost:
...
       enabled: true
       # -- Service name of in-cluster Prometheus
-      serviceName: my-prometheus
+      serviceName: prometheus-server
       # -- Namespace of in-cluster Prometheus
-      namespaceName: opencost
+      namespaceName: prometheus-system
       # -- Service port of in-cluster Prometheus
-      port: 9090
+      port: 80
```
sathieu, line 263: Friendly reminder: This is a breaking change, and major version should have been bumped (with proper release notes).

PR #163, opened 2024-01-02
add debugPort value for tilt improvements

saweber:
This is a nice to have for tilt; adding this value removes the requirement of changing existing values.yaml files for tilt usage. 

Using the existing extraPorts value is insufficent, because helm `--set` does not support appending a value to a array; it will only either override a specific index or the whole array (both have undesirable side effects).

Related PRs:
https://github.com/opencost/opencost-website/pull/189
https://github.com/opencost/opencost/pull/2428


charts/opencost/Chart.yaml
```diff
@@ -9,7 +9,7 @@ keywords:
   - kubecost
   - opencost
   - monitoring
-version: 1.28.0
+version: 1.28.1
```
mattray, line 12: If you could bump this to 1.28.2 I'll merge

mattray (changes requested):

mattray (approved):

PR #174, opened 2024-01-25
feat: `startupProbe` to avoid a long initial delay seconds for liveness probe

a7i:
https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

Kubernetes Docs: 

>The kubelet uses startup probes to know when a container application has started. If such a probe is configured, liveness and readiness probes do not start until it succeeds, making sure those probes don't interfere with the application startup. This can be used to adopt liveness checks on slow starting containers, avoiding them getting killed by the kubelet before they are up and running.

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -62,6 +62,15 @@ spec:
             - containerPort: {{ .Values.opencost.exporter.apiPort }}
               name: http
           resources: {{- toYaml .Values.opencost.exporter.resources | nindent 12 }}
+          {{- if .Values.opencost.exporter.startupProbe.enabled }}
+          startupProbe:
+            httpGet:
+              path: {{ .Values.opencost.exporter.startupProbe.path }}
+              port: {{ .Values.opencost.exporter.apiPort }}
+            initialDelaySeconds: {{ .Values.opencost.exporter.startupProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.opencost.exporter.startupProbe.periodSeconds }}
+            failureThreshold: {{ .Values.opencost.exporter.startupProbe.failureThreshold }}
+          {{- end }}
           {{- if .Values.opencost.exporter.livenessProbe.enabled }}
           livenessProbe:
             httpGet:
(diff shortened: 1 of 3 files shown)


charts/opencost/values.yaml
```diff
@@ -114,16 +114,28 @@ opencost:
...
       # -- Whether probe is enabled
       enabled: true
       # -- Probe path
       path: /healthz
       # -- Number of seconds before probe is initiated
-      initialDelaySeconds: 120
+      initialDelaySeconds: 10
       # -- Probe frequency in seconds
-      periodSeconds: 10
+      periodSeconds: 20
```
a7i, line 138: readiness probe must fail before liveness probe. otherwise container stops receiving traffic and restarts at the same time.

charts/opencost/values.yaml
```diff
@@ -133,7 +145,7 @@ opencost:
       # -- Probe path
       path: /healthz
       # -- Number of seconds before probe is initiated
-      initialDelaySeconds: 120
+      initialDelaySeconds: 10
```
a7i, line 148: ideally you should run readiness as soon as the container comes up

lmello (approved):
we were having issues in our environment, when we reduced the liveness probe timeout. 

we were having issues in our environment, when we reduced the liveness probe timeout. 
This change will make in the worst case scenario kubernetes wait up to 160 seconds for container start, but it will proceed as soon as the container pass the probe. 

Best case scenario if it starts fast it will wait only 10 seconds. 

This allows kubernetes to wait for the container startup in a more efficient way, soon as the container starts, then it benefits of a faster liveness probe and readiness probe. 

It solves the issues we were having in our environment.

charts/opencost/templates/deployment.yaml
```diff
@@ -62,6 +62,15 @@ spec:
             - containerPort: {{ .Values.opencost.exporter.apiPort }}
               name: http
           resources: {{- toYaml .Values.opencost.exporter.resources | nindent 12 }}
+          {{- if .Values.opencost.exporter.startupProbe.enabled }}
+          readinessProbe:
```
mattray, line 66: shouldn't this be a `startupProbe`?

mattray (changes requested):

charts/opencost/templates/deployment.yaml
a7i: good catch! fixed

mattray (approved):

PR #180, opened 2024-03-06
implement plugin support in helm chart

ameijer:
adds plugin support for the OC helm chart. increments chart version to 1.31

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -46,6 +46,17 @@ spec:
       {{- with.Values.opencost.topologySpreadConstraints }}
       topologySpreadConstraints: {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if (and .Values.plugins.enabled .Values.plugins.install.enabled )}}
+      initContainers:
+        - name: plugin-installer
+          image: {{ .Values.plugins.install.fullImageName }}
+          command: ["sh", "/install/install_plugins.sh"]
+          volumeMounts:
+            - name: install-script
+              mountPath: /install
+            - name: plugins-dir
+              mountPath: {{ .Values.plugins.folder }}
+      {{- end }}
       containers:
         - name: {{ include "opencost.fullname" . }}
           {{- if .Values.opencost.exporter.image.fullImageName }}
@@ -93,6 +104,10 @@ spec:
           securityContext: {{- toYaml . | nindent 12 }}
           {{- end }}
           env:
+            - name: LOG_LEVEL
+              value: {{ .Values.loglevel }}
+            - name: CUSTOM_COST_ENABLED
+              value: {{ .Values.plugins.enabled | quote }}
             {{- if .Values.opencost.exporter.apiPort }}
             - name: API_PORT
               value: {{ .Values.opencost.exporter.apiPort | quote }}
@@ -185,8 +200,19 @@ spec:
             - name: {{ $key }}
               value: {{ $value | quote }}
             {{- end }}
-          {{- if or .Values.opencost.exporter.persistence.enabled .Values.opencost.exporter.extraVolumeMounts .Values.opencost.customPricing.enabled .Values.opencost.cloudIntegrationSecret}}
+          {{- if or .Values.plugins.enabled .Values.opencost.exporter.persistence.enabled .Values.opencost.exporter.extraVolumeMounts .Values.opencost.customPricing.enabled .Values.opencost.cloudIntegrationSecret}}
           volumeMounts:
+            {{- if .Values.plugins.enabled }} 
+            - mountPath: /opt/opencost/plugin
+              name: plugins-dir
+              readOnly: false
+            {{- range $key, $config := .Values.plugins.configs }}
+            - mountPath: /opt/opencost/plugin/config/{{$key}}_config.json
+              subPath: {{$key}}_config.json
+              name: plugins-config
+              readOnly: true
+            {{- end }}
+            {{- end }}
             {{- if .Values.opencost.exporter.persistence.enabled }}
             - mountPath: /mnt/export
               name: opencost-export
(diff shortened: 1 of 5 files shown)


charts/opencost/templates/deployment.yaml
```diff
@@ -46,6 +46,17 @@ spec:
       {{- with.Values.opencost.topologySpreadConstraints }}
       topologySpreadConstraints: {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if (and .Values.plugins.enabled .Values.plugins.install.enabled )}}
+      initContainers:
+        - name: plugin-installer
+          image: {{ .Values.plugins.install.fullImageName }}
```
chipzoller, line 52: Any new containers, init or otherwise, should ideally conform to the Pod Security Standards' [restricted profile](https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted).

charts/opencost/templates/install-plugins.yaml
```diff
@@ -0,0 +1,46 @@
...
+        *amd64*) ARCH='amd64';;
+        *aarch64*) ARCH='arm64';;
+        *arm64*) ARCH='arm64';;
+        *)         echo "$UNAME_OUTPUT is unsupported" && exit 1 ;;
+      esac
+
+      {{- if .Values.plugins.version  }}
+      VER={{ .Values.plugins.version | quote}}
+      {{- else }}
+      VER=$(curl --silent -qI https://github.com/opencost/opencost-plugins/releases/latest | awk -F '/' '/^location/ {print  substr($NF, 1, length($NF)-1)}');
```
chipzoller, line 38:
```suggestion
      VER=$(curl --silent -qI https://api.github.com/repos/opencost/opencost-plugins/releases/latest | jq -r '.tag_name');
```

charts/opencost/templates/deployment.yaml
ameijer: I was able to conform to that with a security context for this container

dwbrown2: @ameijer do you see strong reasons to keep this as a standalone container?

ameijer: @dwbrown2, all this container does is download the plugins the user has configured - it is just an init container that dies after it gets the plugins and put them in a local dir for the regular opencost container. The plugins themselves just run in the existing opencost process. I followed a similar paradigm to Sonarqube, which has a bunch of runtime plugins: https://github.com/SonarSource/helm-chart-sonarqube/blob/master/charts/sonarqube/templates/deployment.yaml#L198

ameijer: We chose this approach of dynamically downloading the plugins like this to avoid container bloat. A plugin could be as much as 30-40% of the existing opencost container image size - this allows us to scale to support many plugins by only getting the ones the user actually wants to use, rather than asking them to download what could become a multi-gigabyte container if we were to embed them all right alongside opencost. It's worth re-iterating that this does not increase the number of running containers in an opencost installation, just adds an init container

chipzoller:
Comments to resolve here. Important to add to the initial post that this PR also bumps the Chart version.

charts/opencost/templates/deployment.yaml
dwbrown2: Awesome, appreciate you sharing! Maybe this thread is enough, but may be good to document this.

ameijer: sure, I will find an intuitive spot for this info

ameijer:
thanks chip, addressed all those comments and added note on chart version bump

charts/opencost/templates/install-plugins.yaml
```diff
@@ -0,0 +1,46 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ template "opencost.fullname" . }}-install-plugins
+  labels:
+    app: {{ template "opencost.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+data:
+  install_plugins.sh: |-
```
toscott, line 12: Is there a long term plan to provide an official versioned plugin installer/image with more narrow configuration for plugin installation? Feels like something a little cleaner than a bash script inside the helm chart would be good for updates. 

I forsee companies needing to customize things like the base plugin url (e.g. self hosted artifact repositories). Also feels like some kind of validation on the plugins after download would be good (md5 sum checks)

charts/opencost/templates/install-plugins.yaml
```diff
@@ -0,0 +1,46 @@
...
+      esac
+
+      {{- if .Values.plugins.version  }}
+      VER={{ .Values.plugins.version | quote}}
+      {{- else }}
+      VER=$(curl --silent https://api.github.com/repos/opencost/opencost-plugins/releases/latest | grep ".tag_name" | awk -F\" '{print $4}')
+      {{- end }}
+
+      {{- range $pluginName := .Values.plugins.configs }}
+      curl -fsSLO "https://github.com/opencost/opencost-plugins/releases/download/$VER/datadog.ocplugin.$OS.$ARCH"
```
toscott, line 42: Did we intend to leverage $pluginName instead of hardcoding datadog here?

toscott (commented):
@ameijer  - Had just a couple questions before approving/merging.

charts/opencost/templates/install-plugins.yaml
ameijer: All good points. 

What I like about embedding the script is that it is sort of just there for people to look at and run wherever they want. As long as you run this in a container (maybe a hardened RHEL UBI image, for example) with curl and the basic unix CLI utils available, then this will run without complaint. If we go with another container, then that is an additional image users need to get through compliance/bring into and manage in their artifact repos.

The MD5 and different base URL ideas are definitely great and key in the medium/long term. I will track these as improvements that we can prioritize based on demand, but have been thinking of this as a basic prototype implementation to validate if this is something people will even want/use before we invest more into it. 

How does that sound?

charts/opencost/templates/install-plugins.yaml
ameijer: d'oh! good catch, thanks

charts/opencost/templates/install-plugins.yaml
ameijer: tracking these asks in https://github.com/opencost/opencost-helm-chart/issues/181

toscott (approved):

PR #182, opened 2024-03-08
feat: add ability to set KSM settings via helm chart values

asdfgugus:
I looked deeper into the helm chart and I noticed two things:

1. IMHO the section name `opencostMetrics` does not fit well. Therefore I would suggest to set it to `opencost.metrics.kubeStateMetrics`
2. It is not common in this helm chart to use the `default` function, which is used in the [cost-analyzer-helm-chart](https://github.com/kubecost/cost-analyzer-helm-chart/blob/develop/cost-analyzer/). 

My suggestion in this draft is to set only the settings configured by the user. In other words, not to set default values at all. Example configuration:
```YAML
kubeStateMetrics:
  emitPodAnnotations: ""
  emitNamespaceAnnotations: ""
  emitKsmV1Metrics: "" 
  emitKsmV1MetricsOnly: ""
```

If default values are preferred, I currently see two options:
1. Use the same approach as the cost-analyzer. Example configuration: 
```YAML
kubeStateMetrics: {}
  # emitPodAnnotations: false
  # emitNamespaceAnnotations: false
  # emitKsmV1Metrics: true
  # emitKsmV1MetricsOnly: false
```

2. Set the default values directly in the values file. However, this would result in always setting these values as an environment variable. Example configuration: 
```YAML
kubeStateMetrics:
  emitPodAnnotations: false
  emitNamespaceAnnotations: false
  emitKsmV1Metrics: true
  emitKsmV1MetricsOnly: false
```

I would like to hear your comments!

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.31.0
+version: 1.32.0
 maintainers:
   - name: mattray
   - name: toscott
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -198,6 +198,22 @@ spec:
               value: {{ .Values.opencost.cloudCost.queryWindowDays | quote }}
             - name: CLOUD_COST_RUN_WINDOW_DAYS
               value: {{ .Values.opencost.cloudCost.runWindowDays | quote }}
+            {{- if .Values.opencost.metrics.kubeStateMetrics.emitPodAnnotations }}
+            - name: EMIT_POD_ANNOTATIONS_METRIC
+              value: {{ ( .Values.opencost.metrics.kubeStateMetrics.emitPodAnnotations) | quote }}
+            {{- end }}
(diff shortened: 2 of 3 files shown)


mattray:
I think I like your recommendation to only set the settings configured by the user and `opencost.metrics.kubeStateMetrics` makes sense to me. The PR looks good, but if you want someone else to give feedback we can wait, or I can review it and merge it.

asdfgugus:
Thanks @mattray for your review! I'm glad to hear the PR looks good to you. I'm open to additional feedback, so if you think waiting for someone else's input would be beneficial, we can certainly do that. Otherwise, you can proceed with reviewing and merging the PR. Let me know your preference.

mattray (approved):

mattray:
I'm fine with it, you need to take it off of Draft for a merge. It also needs to have the version bumped to 1.32.0, that's why CI is unhappy.

asdfgugus:
Thanks for the feedback. I changed it accordingly.

mattray (approved):

mattray:
The linter's unhappy with trailing whitespaces. Fix that and I'll merge.

asdfgugus:
Done! 
The issue was here: 
image

mattray (approved):

PR #193, opened 2024-04-04
chore(kubeStateMetrics): make it clear that the keys of kubeStateMetrics only accept booleans for the values

asdfgugus:
Discussed with @mattray. This should provide a better usability.

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -259,14 +259,14 @@ opencost:
 
   metrics:
     kubeStateMetrics:
-      # -- Enable emission of pod annotations
-      emitPodAnnotations: ""
-      # -- Enable emission of namespace annotations
-      emitNamespaceAnnotations: ""
-      # -- Enable emission of KSM v1 metrics
-      emitKsmV1Metrics: ""
-      # -- Enable only emission of KSM v1 metrics that do not exist in KSM 2 by default
-      emitKsmV1MetricsOnly: ""
+      # -- (bool) Enable emission of pod annotations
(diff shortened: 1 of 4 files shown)


mkilchhofer (approved):
👍🎉

Approach inspired by my kyverno PR - i like:
https://github.com/kyverno/kyverno/pull/9532

charts/opencost/values.yaml
```diff
@@ -259,14 +259,14 @@ opencost:
...
     kubeStateMetrics:
-      # -- Enable emission of pod annotations
-      emitPodAnnotations: ""
-      # -- Enable emission of namespace annotations
-      emitNamespaceAnnotations: ""
-      # -- Enable emission of KSM v1 metrics
-      emitKsmV1Metrics: ""
-      # -- Enable only emission of KSM v1 metrics that do not exist in KSM 2 by default
-      emitKsmV1MetricsOnly: ""
+      # -- (bool) Enable emission of pod annotations
```
mattray, line 262: I'm not familiar with the `~` usage, is that doc'd somewhere (for my own curiosity)?

mattray (approved):

charts/opencost/values.yaml
asdfgugus: This is a YAML standard: https://yaml.org/spec/1.2-old/spec.html#id2805071

mkilchhofer: You can also write 
`foo: null` but `foo: ~` is fancier 😆

asdfgugus:
Can we merge?

PR #196, opened 2024-04-10
Option to use existing kubernetes secret 

mpostument:
Hello, with current approach prometheus username and password should be added to values.yml, and helm create secret based on those values. But in case when secret already exist in kubernetes there is no way to use it with opencost helm cahrt. This PR allowing to use existing secrets

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.33.3
+version: 1.34.0
 maintainers:
   - name: mattray
     url: https://mattray.dev
--- a/charts/opencost/README.md
+++ b/charts/opencost/README.md
@@ -117,7 +117,7 @@ $ helm install opencost opencost/opencost
 | opencost.prometheus.amp.workspaceId | string | `""` | Workspace ID for AMP |
 | opencost.prometheus.bearer_token | string | `""` | Prometheus Bearer token |
 | opencost.prometheus.bearer_token_key | string | `"DB_BEARER_TOKEN"` |  |
-| opencost.prometheus.existing_secret_name | string | `nil` | Existing secret name that contains credentials for Prometheus |
+| opencost.prometheus.existingSecretName | string | `nil` | Existing secret name that contains credentials for Prometheus |
 | opencost.prometheus.external.enabled | bool | `false` | Use external Prometheus (eg. Grafana Cloud) |
 | opencost.prometheus.external.url | string | `"https://prometheus.example.com/prometheus"` | External Prometheus url |
 | opencost.prometheus.internal.enabled | bool | `true` | Use in-cluster Prometheus |
--- a/charts/opencost/README.md
+++ b/charts/opencost/README.md
@@ -2,7 +2,7 @@
 
 OpenCost and OpenCost UI
 
-![Version: 1.33.3](https://img.shields.io/badge/Version-1.33.3-informational?style=flat-square)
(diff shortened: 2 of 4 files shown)


charts/opencost/Chart.yaml
```diff
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.33.3
+version: 1.33.4
```
asdfgugus, line 12: This is a new feature. Please bump to 1.34.0.

mattray (changes requested):
If you could bump the version to 1.34.0, we'll merge.

charts/opencost/templates/deployment.yaml
```diff
@@ -167,6 +167,27 @@ spec:
                   name: {{ include "opencost.prometheus.secretname" . }}
                   key: {{ .Values.opencost.prometheus.bearer_token_key }}
             {{- end }}
+            {{- if and .Values.opencost.prometheus.existing_secret_name .Values.opencost.prometheus.username_key }}
+            # If username or password, already exist in secret
+            - name: DB_BASIC_AUTH_USERNAME
```
asdfgugus, line 172: This part is now duplicate. What about checking if `.Values.opencost.prometheus.existing_secret_name` is set and else if `.Values.opencost.prometheus.secret_name` is set?

mpostument:
> If you could bump the version to 1.34.0, we'll merge.

Done

charts/opencost/templates/deployment.yaml
mpostument: Not sure i understand logic. Can you share example how you propose to do it?

asdfgugus: Without testing every case, I thought about something like this: 
```YAML
{{- if and .Values.opencost.prometheus.username_key (or .Values.opencost.prometheus.username .Values.opencost.prometheus.existingSecretName) }}
- name: DB_BASIC_AUTH_USERNAME
  valueFrom:
    secretKeyRef:
      name: {{ .Values.opencost.prometheus.existingSecretName | default (include "opencost.prometheus.secretname" .) }}
      key: {{ .Values.opencost.prometheus.username_key }}
{{- end }}
```
---
IMHO the key for the existing secret name should be camelCase. But I saw some other snake_case keys in the code.

asdfgugus:
@mpostument can you combine the two commits for a better traceability?

charts/opencost/templates/deployment.yaml
mpostument: Make sense, will do it

mpostument: Done

mpostument:
> @mpostument can you combine the two commits for a better traceability?

Done

charts/opencost/README.md
```diff
@@ -117,6 +117,7 @@ $ helm install opencost opencost/opencost
 | opencost.prometheus.amp.workspaceId | string | `""` | Workspace ID for AMP |
 | opencost.prometheus.bearer_token | string | `""` | Prometheus Bearer token |
 | opencost.prometheus.bearer_token_key | string | `"DB_BEARER_TOKEN"` |  |
+| opencost.prometheus.existing_secret_name | string | `nil` | Existing secret name that contains credentials for Prometheus |
```
asdfgugus, line 120: run helm-docs again

mpostument: done

asdfgugus (approved):
LGTM!

mattray (approved):

PR #202, opened 2024-05-03
Option to support additional Service types

dbcelm:
Greetings..!!

I have added option to choose between ClusterIP, NodePort and LoadBalancer service types. This will come handy for variety of use cases:

e.g. when using aws-load-balancer controller with Cilium CNI default ClusterIP service type won't work and will give 504 when being reached from ingress as ALB Target Group won't be able to send health check traffic to ServiceIP as it will be Cilium managed and not from VPC CIDR range, hence user has to opt for target-type as instance and hence only Service Type NodePort will work in that case.

Also added test case for service template.


asdfgugus (commented):
Thanks for the PR! 
I added some comments. Could a maintainer verify them? Thanks! 
In addition, please run the helm-docs command.

dbcelm:
@asdfgugus pushed changes as per comments and also ran helm-docs

charts/opencost/templates/service.yaml
```diff
@@ -14,7 +14,17 @@ metadata:
   {{- end }}
 spec:
   selector: {{- include "opencost.selectorLabels" . | nindent 4 }}
-  type: {{ .Values.service.type }}
+{{- if and .Values.service .Values.service.type }}
+  type: "{{ .Values.service.type }}"
+{{- else }}
+  type: ClusterIP
```
asdfgugus, line 20: Already default.

charts/opencost/templates/service.yaml
```diff
@@ -14,7 +14,17 @@ metadata:
...
+{{- end }}
+{{- if (eq .Values.service.type "LoadBalancer") }}
+{{- if .Values.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges:
+{{ toYaml .Values.service.loadBalancerSourceRanges | indent 4 }}
+  {{- end -}}
```
asdfgugus, line 26: Indentation does not match.

dbcelm:
@asdfgugus made required edits

asdfgugus:
Sorry, I meant to remove the two spaces. 
To finalize, could you merge the commits together for a better visibility?

dbcelm:
> Sorry, I meant to remove the two spaces. To finalize, could you merge the commits together for a better visibility?
 if for "loadBalancerSourceRanges" then its desired indent with type:
 
 merged last 2 commits

mattray:
This looks good to me, could you bump the chart version to 1.36.0?

dbcelm:
> This looks good to me, could you bump the chart version to 1.36.0?

bumped chart version to 1.36.0

mattray:
@dbcelm sorry, there's been another release. Could you move it to 1.37.0?

dbcelm:
> @dbcelm sorry, there's been another release. Could you move it to 1.37.0?

done

mattray (approved):

Issue #205, opened 2024-05-15
Failing when customPricing set 


mogaal:
Using the latest version of the Helm Chart (`version: 1.37.0`) and setting customPricing to true (`.Values.opencost.customPricing.enabled`) cause the pod to fail with the following error:

```
opencost-ui /docker-entrypoint.sh: Configuration complete; ready for start up
opencost-ui 2024/05/15 10:27:58 [notice] 1#1: using the "epoll" event method
opencost-ui 2024/05/15 10:27:58 [notice] 1#1: nginx/1.25.5
opencost-ui 2024/05/15 10:27:58 [notice] 1#1: built by gcc 13.2.1 20231014 (Alpine 13.2.1_git20231014) 
opencost-ui 2024/05/15 10:27:58 [notice] 1#1: OS: Linux 6.6.22-linuxkit
opencost-ui 2024/05/15 10:27:58 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
opencost-ui 2024/05/15 10:27:58 [notice] 1#1: start worker processes
opencost 2024-05-15T10:27:58.127117389Z INF No metrics-config configmap found at install time, using existing configs: configmaps "metrics-config" not found
opencost 2024-05-15T10:27:58.127243847Z INF Failed to get metrics config before init: error reading metrics config file: open /tmp/custom-config/metrics.json: not a directory
opencost 2024-05-15T10:27:58.127567764Z WRN Controller: pullWatchers: failed to load config statuses from file: open /tmp/custom-config/cloud-configurations.json: not a directory. Proceeding to create the file
opencost 2024-05-15T10:27:58.127651722Z ERR Controller: pullWatchers: failed to save statuses failed to save config statuses to file: open /tmp/custom-config/cloud-configurations.json: not a directory
opencost 2024-05-15T10:27:58.127671222Z ERR MultiCloudWatcher:  error checking file at '/tmp/custom-config/cloud-integration/cloud-integration.json': stat /tmp/custom-config/cloud-integration/cloud-integration.json: not a directory
opencost 2024-05-15T10:27:58.127685347Z INF Custom Pricing file at path '/tmp/custom-config/default.json' read error: 'stat /tmp/custom-config/default.json: not a directory'
opencost 2024-05-15T10:27:58.127701597Z INF Custom Pricing file at path '/tmp/custom-config/default.json' read error: 'stat /tmp/custom-config/default.json: not a directory'
[...]

PR #206, opened 2024-05-15
Fix error when loading custom pricing configuration file 

mogaal:
This fixes the bug I reported in https://github.com/opencost/opencost-helm-chart/issues/205, where the `CONFIG_PATH` directory is mounted as a file instead of as a directory:

```
~ $ ls /tmp/custom-config -l
-rw-r--r--    1 root     root           299 May 15 10:27 /tmp/custom-config
~ $ cat /tmp/custom-config 
{
  "CPU": "1.25",
  "GPU": "0.95",
  "RAM": "0.5",
  "description": "Modified pricing configuration.",
  "internetNetworkEgress": "0.12",
  "regionNetworkEgress": "0.01",
  "spotCPU": "0.006655",
  "spotRAM": "0.000892",
  "storage": "0.25",
  "zoneNetworkEgress": "0.01",
  "provider" : "custom"
}~ $
```

This causes the pod to fail with:

```
opencost 2024-05-15T10:27:58.127897889Z ERR error accessing cloud provider configuration: stat /tmp/custom-config/default.json: not a directory
```

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -243,7 +243,7 @@ spec:
               readOnly: false
             {{- end }}
             {{- if .Values.opencost.customPricing.enabled }}
-            - mountPath: {{ .Values.opencost.customPricing.configPath }}
+            - mountPath: {{ .Values.opencost.customPricing.configPath }}/{{ include "opencost.configFileName" . }}.json
               name: custom-configs
               subPath: {{ include "opencost.configFileName" . }}.json
               readOnly: true
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.37.0
+version: 1.38.0
 maintainers:
   - name: mattray
     url: https://mattray.dev


mogaal:
@mattray Hi, when you have some time, pls let me know if this looks good to you 🙏

charts/opencost/templates/deployment.yaml
```diff
@@ -243,7 +243,7 @@ spec:
               readOnly: false
             {{- end }}
             {{- if .Values.opencost.customPricing.enabled }}
-            - mountPath: {{ .Values.opencost.customPricing.configPath }}
+            - mountPath: {{ .Values.opencost.customPricing.configPath }}/{{ include "opencost.configFileName" . }}.json
               name: custom-configs
               subPath: {{ include "opencost.configFileName" . }}.json
```
mattray, line 248: the subPath isn't working?

mogaal: No, it gets mounted as a file:

```
~ $ ls /tmp/custom-config -l
-rw-r--r--    1 root     root           299 May 15 10:27 /tmp/custom-config
~ $ cat /tmp/custom-config 
{
  "CPU": "1.25",
  "GPU": "0.95",
  "RAM": "0.5",
  "description": "Modified pricing configuration.",
  "internetNetworkEgress": "0.12",
  "regionNetworkEgress": "0.01",
  "spotCPU": "0.006655",
  "spotRAM": "0.000892",
  "storage": "0.25",
  "zoneNetworkEgress": "0.01",
  "provider" : "custom"
}~ $
```

Riyaz-Behrad: As @mogaal said,
The subPath specifies what key should be loaded from the configmap. The mountPath is the full path to the file loaded, so it should also contain the file's name or else the custom-config would be considered as the mounted file's name. Final applied deployment should be like this:
```
        - mountPath: /tmp/custom-config/default.json
          name: custom-configs
          readOnly: true
          subPath: default.json
```

valdisrigdon:
https://github.com/opencost/opencost-helm-chart/pull/208 should have the full fix for this, as you need to fix the custom metrics too.

mattray (approved):

PR #207, opened 2024-05-27
fix: namspace label key in selectors for netpol

y-eight:
Currently, the namespace label selector in the NetworkPolicies that will be created by setting the value `networkPolicies.enabled` to `true` is not set properly. The key of the namespace selector in the egress and ingress definition is `name` but should be `kubernetes.io/metadata.name`.
The [kubenetes docs](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name) state that the label `kubernetes.io/metadata.name` is set automatically and the `name` label is not compulsorily existing.


mattray (approved):

mattray:
Needs a minor version bump, I'll do it when I test next

PR #212, opened 2024-06-14
Bump all the GitHub Action dependencies

mattray:
GHA were failing on publish, let's see if this straightens it out.


toscott (approved):
Assuming the builds are still functioning, should be fine imo.

PR #215, opened 2024-07-01
feat: include image digest to prevent supply chain issues

### LINKED ISSUE #214: Include image digest for containers
Proposal
Explicitly set the image digests in the helm chart, while still allowing users to remove them if desired. 

E.G.
```
opencost:
  exporter:
    image:
      digest: sha256:6aa68e52a24b14ba41f23db08d1b9db1429a1c0300f4c0381ecc2c61fc311a97
```

Why
Without the digest, anyone pulling images directly from the public registry could be at risk of supply chain problems. While existing images tags should not be getting overwritten, without an image digest in the helm chart any repush will automatically propagate out to those users.

toscott:
fixes #214

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.40.0
+version: 1.41.0
 maintainers:
   - name: mattray
     url: https://mattray.dev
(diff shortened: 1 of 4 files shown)


toscott:
Opted to include it in the tag over making a separate digest to keep the image tag configuration in one spot and preserve backward compatibility with anyone already explicitly overriding the image tag.

mattray (approved):
Works with my local testing for both arm64 and amd64 with the SHA.

PR #218, opened 2024-07-25
Allow other forms of extra env variables to allow secrets

### LINKED ISSUE #209: Allow for an existing cloudProviderApiKey secret
The helm chart allows for the use of an existing prometheus secret

https://github.com/opencost/opencost-helm-chart/blob/c86ac7031c2965c839a757591a73d3f32f2a9e93/charts/opencost/templates/deployment.yaml#L149

However an existing cloudProviderApiKey secret cannot be specified.

https://github.com/opencost/opencost-helm-chart/blob/c86ac7031c2965c839a757591a73d3f32f2a9e93/charts/opencost/templates/deployment.yaml#L126C46-L126C65

Sadzeih:
This should resolve #209

### DIFF
--- a/charts/opencost/README.md
+++ b/charts/opencost/README.md
@@ -64,7 +64,7 @@ $ helm install opencost opencost/opencost
 | opencost.exporter.defaultClusterId | string | `"default-cluster"` | Default cluster ID to use if cluster_id is not set in Prometheus metrics. |
 | opencost.exporter.env | list | `[]` | List of additional environment variables to set in the container |
 | opencost.exporter.extraArgs | list | `[]` | List of extra arguments for the command, e.g.: log-format=json |
-| opencost.exporter.extraEnv | object | `{}` | Any extra environment variables you would like to pass on to the pod |
+| opencost.exporter.extraEnv | list | `[]` | A list of extra environment variables you would like to pass on to the pod |
 | opencost.exporter.extraVolumeMounts | list | `[]` | A list of volume mounts to be added to the pod |
 | opencost.exporter.image.fullImageName | string | `nil` | Override the full image name for development purposes |
 | opencost.exporter.image.pullPolicy | string | `"IfNotPresent"` | Exporter container image pull policy |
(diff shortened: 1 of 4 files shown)


mattray (approved):

mattray:
I don't have cycles to test it, but the patch looks good. If someone else gives it a thumbs up I'll merge and release.

charts/opencost/README.md
```diff
@@ -64,7 +64,7 @@ $ helm install opencost opencost/opencost
 | opencost.exporter.defaultClusterId | string | `"default-cluster"` | Default cluster ID to use if cluster_id is not set in Prometheus metrics. |
 | opencost.exporter.env | list | `[]` | List of additional environment variables to set in the container |
```
toscott, line 65: Does this existing variable (env instead of extraEnv) not provide whats needed here?

charts/opencost/README.md
```diff
@@ -64,7 +64,7 @@ $ helm install opencost opencost/opencost
...
 | opencost.exporter.extraArgs | list | `[]` | List of extra arguments for the command, e.g.: log-format=json |
-| opencost.exporter.extraEnv | object | `{}` | Any extra environment variables you would like to pass on to the pod |
+| opencost.exporter.extraEnv | list | `[]` | A list of extra environment variables you would like to pass on to the pod |
```
toscott, line 67: Swapping the type of object is backwards incompatible. For the sanity of downstream users, I recommend we  keep breaking changes to only those absolutely required, and when we do we increment the chart major version.

toscott (commented):
Given there is an existing property that has this behavior and this is a breaking change, going to decline this one. Please let me know if I am misunderstanding something here..

charts/opencost/README.md
Sadzeih: That's definitely a misread on my part, didn't realize there was two ways of adding environment variables, and since the "standard" is `extraEnv` in most helm charts, I didn't look further. Thanks!

PR #222, opened 2024-08-13
Update _helpers.tpl

thunderthome:
Make sure that prometheus internal enabled really works (that i can disable it).

### DIFF
--- a/charts/opencost/templates/_helpers.tpl
+++ b/charts/opencost/templates/_helpers.tpl
@@ -98,7 +98,7 @@ Create the name of the controller service account to use
     {{- $port := .Values.opencost.sigV4Proxy.port | int }}
     {{- $ws := .Values.opencost.prometheus.amp.workspaceId }}
     {{- printf "http://localhost:%d/workspaces/%v" $port $ws -}}
-  {{- else -}}
+  {{- else if .Values.opencost.prometheus.internal.enabled -}}
     {{- $host := tpl .Values.opencost.prometheus.internal.serviceName . }}
     {{- $ns := tpl .Values.opencost.prometheus.internal.namespaceName . }}
     {{- $port := .Values.opencost.prometheus.internal.port | int }}


toscott:
Hi @thunderthome. Can you help me understand the use case here? 

Opencost requires a prometheus endpoint. The options are basically an external one or an internal one. If its internal, the network policy gets added (which doesn't hurt to add if its an external prometheus, so its just defaulted on.

thunderthome:
@toscott the reason I want to add this is because now if I turn on Thanos my values aren't used because it has the fallback in the prometheus setup. I wanted to use it internally at first, but one of my colleagues pointed to me that we use Thanos. Now if I want to enable Thanos and disable internal Prometheus it won't let me do that because of this fallback.

mittal-ishaan:
> now if I turn on Thanos my values aren't used because it has the fallback in the prometheus setup.

@thunderthome could you help me understand this thing. what fallback are you referring to here?

thunderthome:
@mittal-ishaan if the statement remains else, it will always take that part of the values of prometheus, even if it's disabled. Now that it checks if it's enabled or disabled it can grab it or not. If not then it will continue to the Thanos part. 

The reason why it was probably setup like this in the first place is because either one of them needs to be configured for it to work. So it was probably a fallback solution.

charts/opencost/templates/_helpers.tpl
```diff
@@ -98,7 +98,7 @@ Create the name of the controller service account to use
     {{- $port := .Values.opencost.sigV4Proxy.port | int }}
     {{- $ws := .Values.opencost.prometheus.amp.workspaceId }}
     {{- printf "http://localhost:%d/workspaces/%v" $port $ws -}}
-  {{- else -}}
+  {{- else if .Values.opencost.prometheus.internal.enabled -}}
```
ameijer, line 101: if we default this to true in values, maybe that is fail safe enough?

ameijer: @thunderthome would you be able to set values.yaml to have this default?

PR #229, opened 2024-10-06
Add kubeRBACProxy property to support querying in cluster prometheus in openshift

mittal-ishaan:
* Add `kubeRBACProxy` property to set `KUBE_RBAC_PROXY_ENABLED` as true in the env var of the container.
* `createPrometheusClusterRoleBinding` property to create cluster role binding to grant required permissions to the serviceaccount to query prometheus with [kube-rbac-proxy](https://catalog.redhat.com/software/containers/rhacm2/kube-rbac-proxy-rhel8/611d4320eb34b73652492bb8) enabled. 
* Add the role binding to grant permissions this external prometheus to scrape metrics from kubecost. 
* Add fail if both `kubeRBACProxy` and `BearerToken` property cannot be set together.

Testing:
deployed opencost in openshift cluster using in cluster prometheus. We were able to query the prometheus.
```sh
2024-12-06T19:03:40.832625335Z INF Prometheus/Thanos Client Max Concurrency set to 5                                                                                                       2024-12-06T19:03:41.881115203Z INF Success: retrieved the 'up' query against prometheus at: https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091                            
2024-12-06T19:03:41.882834227Z INF Retrieved a prometheus config file from: https://prometheus-k8s.openshift-monitoring.svc.cluster.local:9091                                             
2024-12-06T19:03:41.955015089Z INF Using scrape interval of 60.000000     
```

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.42.4
+version: 1.43.0
 maintainers:
   - name: mattray
     url: https://mattray.dev
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.42.3
+version: 1.42.4
 maintainers:
   - name: mattray
     url: https://mattray.dev
--- a/charts/opencost/templates/_helpers.tpl
+++ b/charts/opencost/templates/_helpers.tpl
@@ -122,10 +122,10 @@ Check that either thanos external or internal is defined
 
 {{/*
   Fail if both kube-rbac-proxy and bearer token are set
-}}
+*/}}
 {{- define "kubeRBACProxyBearerTokenCheck" -}}
-{{- if and (.Values.opencost.prometheus.kubeRBACProxy .Values.opencost.prometheus.bearer_token) }}
-  {{- fail "\n\nBoth kubeRBACProxy and bearer_token are set. Please specify only one." -}}
+{{- if and .Values.opencost.prometheus.kubeRBACProxy .Values.opencost.prometheus.bearer_token }}
+  {{- fail "Both kubeRBACProxy and bearer_token are set. Please specify only one." -}}
 {{- end -}}
 {{- end -}}
 
--- a/charts/opencost/templates/_helpers.tpl
+++ b/charts/opencost/templates/_helpers.tpl
@@ -121,7 +121,7 @@ Check that either thanos external or internal is defined
 {{- end -}}
 
 {{/*
-  Verify if both kube-rbac-proxy and bearer token are set
+  Fail if both kube-rbac-proxy and bearer token are set
 }}
 {{- define "kubeRBACProxyBearerTokenCheck" -}}
 {{- if and (.Values.opencost.prometheus.kubeRBACProxy .Values.opencost.prometheus.bearer_token) }}
--- a/charts/opencost/templates/_helpers.tpl
+++ b/charts/opencost/templates/_helpers.tpl
@@ -121,6 +121,15 @@ Check that either thanos external or internal is defined
 {{- end -}}
 
 {{/*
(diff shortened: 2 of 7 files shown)


kastl-ars:
Any news on this? This would be really nice to have, as I have a client with OpenShift which is evaluating OpenCost at the moment...

mittal-ishaan:
Hi @kastl-ars let me pick this up this week. Got busy with other tasks

charts/opencost/templates/_helpers.tpl
```diff
@@ -120,6 +120,15 @@ Check that either thanos external or internal is defined
   {{- end -}}
 {{- end -}}
 
+{{/*
+  Verify if both kube-rbac-proxy and bearer token are set
```
kastl-ars, line 124: Shouldn't this read `Fail if both ... are set`?

charts/opencost/templates/clusterrolebinding.yaml
```diff
@@ -15,4 +15,27 @@ subjects:
...
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "opencost.fullname" . }}-operator
+  labels: {{- include "opencost.labels" . | nindent 4 }}
+  {{- with .Values.annotations }}
+  annotations: {{- toYaml . | nindent 4 }}
+  {{- end }}
+roleRef:
+  # Grant the kubecost service account the cluster-monitoring-view role to enable it to query OpenShift Prometheus.
+  # This is necessary for Kubecost to get access and query the in-cluster Prometheus instance using its service account token.
```
thomasvn, line 31:
```suggestion
  # Grant the OpenCost ServiceAccount the cluster-monitoring-view role to enable it to query OpenShift Prometheus.
  # This is necessary for OpenCost to get access and query the in-cluster Prometheus instance using its service account token.
```

charts/opencost/values.yaml
```diff
@@ -333,6 +333,17 @@ opencost:
     # -- Prometheus Bearer token
     bearer_token: ""
     bearer_token_key: DB_BEARER_TOKEN
+    # -- If true, opencost will use kube-rbac-proxy to authenticate with in cluster Prometheus for openshift
+    kubeRBACProxy: false
+    # OPTIONAL. The following configs only to be enabled when using a Prometheus instance already installed in the cluster.
+    # -- If true, opencost will create a ClusterRoleBinding to grant the Kubecost serviceaccount access to query Prometheus.
```
thomasvn, line 339:
```suggestion
    # -- If true, opencost will create a ClusterRoleBinding to grant the OpenCost ServiceAccount access to query Prometheus.
```

kastl-ars: Shouldn't this read `If true, the helm chart will create...`?

charts/opencost/templates/clusterrolebinding.yaml
kastl-ars: `to query the OpenShift Prometheus`?

thomasvn: These changes were originally made to support querying the default Prometheus running in OpenShift. However, I agree that this could be generalized to any Prometheus requiring `KUBE_RBAC_PROXY`!

I've edited my comment above to generalize.

charts/opencost/values.yaml
thomasvn: Good callout, I'll update my comment above!

toscott (commented):
This will need a version bump. Otherwise lgtm.

thomasvn (approved):

charts/opencost/Chart.yaml
```diff
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.42.3
+version: 1.42.4
```
thomasvn, line 12: Actually, upon further review of all the releases shown here https://github.com/opencost/opencost-helm-chart/releases, this PR makes changes which are significant enough that are probably a bump in minor version not just patch version

mittal-ishaan: Sure, bumping it to `1.43.0`

cliffcolvin (approved):

PR #239, opened 2024-12-06
allow creating an OpenShift route

kastl-ars:
This is **not** meant as a PR ready to merge. I just wanted to start a discussion, if the functionality to create an OpenShift route via the helm chart would be something worth adding.

I know the chart can create an ingress, but if you want to put OpenCost behind a oauth-proxy, you need to put the route name into an annotation on the `serviceAccount`. And if the route is being created from an ingress automatically, its name contains a random hash...

### DIFF
--- a/charts/opencost/templates/route.yaml
+++ b/charts/opencost/templates/route.yaml
@@ -1,4 +1,4 @@
-{{- if and .Capabilities.APIVersions.Has "route.openshift.io/v1/Route" .Values.opencost.ui.enabled .Values.opencost.ui.route.enabled }}
+{{- if and (.Capabilities.APIVersions.Has "route.openshift.io/v1/Route") .Values.opencost.ui.enabled .Values.opencost.ui.route.enabled }}
 apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
--- a/charts/opencost/templates/route.yaml
+++ b/charts/opencost/templates/route.yaml
@@ -1,0 +1,32 @@
+{{- if and .Values.opencost.ui.enabled .Values.opencost.ui.route.enabled }}
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: {{ include "opencost.fullname" . }}-route
+  namespace: {{ include "opencost.namespace" . }}
+  labels: {{- include "opencost.labels" . | nindent 4 }}
+  {{- with .Values.opencost.ui.route.annotations }}
+  annotations: {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  host: {{ .Values.opencost.ui.route.host | quote }}
+  {{- if .Values.opencost.ui.route.path }}
+  path: {{ .Values.opencost.ui.route.path }}
+  {{- end }}
+  port:
+    targetPort: {{ .Values.opencost.ui.route.targetPort }}
+  to:
+    kind: Service
+    name: opencost
+    weight: 100
+  wildcardPolicy: None
+  {{- if .Values.opencost.ui.route.tls }}
+  tls:
+    {{- if .Values.opencost.ui.route.tls.insecureEdgeTerminationPolicy }}
+    insecureEdgeTerminationPolicy: {{ .Values.opencost.ui.route.tls.insecureEdgeTerminationPolicy }}
+    {{- end }}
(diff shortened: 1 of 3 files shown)


AjayTripathy:
@kastl-ars I'm supportive of this idea! @mittal-ishaan you've been working in this area, do you have any thoughts?

mittal-ishaan:
I completely agree with @kastl-ars
In addition, I’d like to propose a discussion around introducing an `openshift` block in the Helm chart to have all OpenShift-specific configurations, such as Routes and in-cluster Prometheus integrations. This change would:

- Ensure OpenShift-specific features are enabled only in OpenShift environments.
- Address specific requirements like volume mounts for OpenShift, avoiding workarounds such as those mentioned in [opencost-ui issue #4](https://github.com/opencost/opencost-ui/issues/4). This could either be solved or a workaround be included as a default in helm chart itself. 

Thoughts?

kastl-ars:
I think the addition of an `openshift` block, that globally triggers e.g. the proper setting of the securityContext, would be nice. But I would propose to keep things separate, i.e. let this PR only be about adding a route.

Otherwise I fear this becomes to big to properly test and review...

mittal-ishaan:
yes, agree we can do it in a separate PR.

kastl-ars:
Does anyone have better documentation on OpenShift routes than the incomplete and "going into obscure corner cases too fast" documentation [here](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html-single/networking/index#nw-creating-a-route_route-configuration). Especially the exact details on the `tls` section are hard to find...

mittal-ishaan:
@kastl-ars any update here? If you want to continue on this

kastl-ars:
> @kastl-ars any update here? If you want to continue on this

Sorry, I lost track of this due to the other issues. I just rebased and will have a look at the CI issues.

kastl-ars:
CI is green again, the PR was missing a version bump.

I currently do not have a OpenShift Cluster available where I can test this. Running `helm template` looks correct, though.

Anything missing? Comments? Improvements?

kastl-ars:
Anything missing on this one? The checks are green...

charts/opencost/templates/route.yaml
```diff
@@ -0,0 +1,32 @@
...
+spec:
+  host: {{ .Values.opencost.ui.route.host | quote }}
+  {{- if .Values.opencost.ui.route.path }}
+  path: {{ .Values.opencost.ui.route.path }}
+  {{- end }}
+  port:
+    targetPort: {{ .Values.opencost.ui.route.targetPort }}
+  to:
+    kind: Service
+    name: opencost
```
mittal-ishaan, line 20:
```suggestion
    name: {{ include "opencost.fullname" . }}
```
https://github.com/opencost/opencost-helm-chart/blob/c810b378802b24beaae0480f31b97be7beaca88a/charts/opencost/templates/service.yaml#L5

charts/opencost/templates/route.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if and .Values.opencost.ui.enabled .Values.opencost.ui.route.enabled }}
```
mittal-ishaan, line 1:
```suggestion
{{- if and .Capabilities.APIVersions.Has "route.openshift.io/v1/Route" .Values.opencost.ui.enabled .Values.opencost.ui.route.enabled }}
```
Just to ensure that non-openshift users do not install it.
https://helm.sh/docs/chart_template_guide/builtin_objects/

mittal-ishaan: not that much important though. the helm values clearly states that it is `openshift route` and disabled by default

charts/opencost/templates/route.yaml
kastl-ars: Fixed.

charts/opencost/templates/route.yaml
kastl-ars: Nevertheless added this.

kastl-ars (commented):
Pushed a new commit with the suggested fixes

mittal-ishaan (approved):

mittal-ishaan (approved):
lgtm
Thank you @kastl-ars

PR #242, opened 2024-12-20
clusterrole.yaml: add if condition for .Values.rbac.enabled

### LINKED ISSUE #241: Creation of ClusterRole and Clusterrolebindings cannot be disabled
Dear maintainers,

as far as I can see, the `rbac.enabled` value does not fully work. The ClusterRole does not have any if-condition pointing to `.Values.rbac.enabled`:

https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/templates/clusterrole.yaml

The `ClusterRoleBinding` correctly has a `if .Values.rbac.enabled` condition set:

https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/templates/clusterrolebinding.yaml#L1

Kind Regards,
Johannes

kastl-ars:
clusterrole.yaml: add if condition for .Values.rbac.enabled, like on the clusterrolebinding

fix #241

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 1.43.0
+version: 1.43.1
 maintainers:
   - name: mattray
     url: https://mattray.dev
--- a/charts/opencost/templates/clusterrole.yaml
+++ b/charts/opencost/templates/clusterrole.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.rbac.enabled }}
 # Cluster role giving opencost to get, list, watch required resources
 # No write permissions are required
 apiVersion: rbac.authorization.k8s.io/v1
@@ -77,3 +78,4 @@ rules:
       - get
       - list
       - watch
+{{- end }}
--- a/charts/opencost/templates/clusterrolebinding.yaml
+++ b/charts/opencost/templates/clusterrolebinding.yaml
@@ -37,5 +37,5 @@ subjects:
   - kind: ServiceAccount
     name: {{ template "opencost.serviceAccountName" . }}
     namespace: {{ include "opencost.namespace" . }}
-{{- end }}
 ---
+{{- end }}


mittal-ishaan:
Hi @kastl-ars 
Thank you for this. LGTM
Due to my recent [merge](https://github.com/opencost/opencost-helm-chart/pull/229), we did add rbac like this which we would want to consider upon. My bad here, I should have seen it under my PR. 
Should we also have them under this if condition too, we are already gating them under a helm config?

Question is do we also want to have `rbac.enabled` flag gate the cluster role needed in openshift environment too (that looks exactly what this flag is intended to do) or we can have scenarios where a user wants to disable the `rbac.enabled` flag but want the required cluster role, etc for openshift environment.

Otherwise the Issue raised and PR is really helpful

kastl-ars:
> Hi @kastl-ars Thank you for this. LGTM Due to my recent [merge](https://github.com/opencost/opencost-helm-chart/pull/229), we did add rbac like this which we would want to consider upon. My bad here, I should have seen it under my PR. Should we also have them under this if condition too, we are already gating them under a helm config?

Sorry, I am not quite sure what you mean?

> Question is do we also want to have `rbac.enabled` flag gate the cluster role needed in openshift environment too (that looks exactly what this flag is intended to do) or we can have scenarios where a user wants to disable the `rbac.enabled` flag but want the required cluster role, etc for openshift environment.

In my opinion, as the OpenShift stuff needs to be enabled explicitly and there is already a value to enable/disable the creation of the clusterRoleBinding, I would say they can be left independent of each other.

https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/values.yaml#L340
https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/values.yaml#L342

The cluster I want to run OpenCost in does not allow me to create roles and rolebindings, so I will keep these settings disabled and have the cluster owner create them for me. But I will enable the general OpenShift integration.

kastl-ars:
Is there anything from blocking this from being merged and released? This prevents me from running OpenCost in one of my clusters, where my permissions do not allow creating roles and rolebindings. And the fix is simple...

kastl-ars:
Could someone have a look and merge this, please? It's really simple...

mittal-ishaan (approved):

cliffcolvin (approved):

mittal-ishaan:
This will need a version bump. Otherwise all fine

cliffcolvin:
@mittal-ishaan you want to bump that rq and I'll approve and merge?

mittal-ishaan:
I am not sure if I might be able to add a commit in the PR as it is a fork. Should I go ahead creating a seperate PR with this change and version bump to get this in quickly?

kastl-ars:
Hi all, I just pushed a commit that changes the chart version to 1.43.1 (or should this be 1.44.0?).

Anything else missing?

Not sure about the `---` markers in charts/opencost/templates/clusterrolebinding.yaml, shouldn't they be inside the `if` condition, too?

mittal-ishaan:
version bump is fine. I think the last `---` maker outside the if condition in that file is not even necessary there and can be removed. I will remove it in another PR. or you can include that change in this PR too.

kastl-ars:
> version bump is fine. I think the last `---` maker outside the if condition in that file is not even necessary there and can be removed. I will remove it in another PR. or you can include that change in this PR too.

I just pushed a commit that moves it into the if condition. Or should I remove it completely?

PR #244, opened 2024-12-26
mount nginx.conf as a volume to the frontend container

mittal-ishaan:
mount `nginx.conf` as a volume to the frontend container. This way we would me moving out of [mounting](https://github.com/opencost/opencost-ui/blob/aa7fbcbc1ad1789ef924851e1128514c1a5bc44d/docker-entrypoint.sh#L22) the nginx.conf to the container when pulling the image in the container as that has permission issues in openshift clusters.

Discussion: https://github.com/opencost/opencost-ui/pull/49#issuecomment-2527030430

We need to bump the `opencost-ui` image to have this https://github.com/opencost/opencost-ui/pull/49 change

Testing: I was able to install opencost with these changes


charts/opencost/templates/deployment.yaml
```diff
@@ -312,8 +312,14 @@ spec:
           {{- with .Values.opencost.ui.securityContext }}
           securityContext: {{- toYaml . | nindent 12 }}
           {{- end }}
+          volumeMounts:
           {{- with .Values.opencost.ui.extraVolumeMounts }}
-          volumeMounts: {{- toYaml . | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- if .Values.opencost.ui.enabled }}
```
toscott, line 319: This whole section (270 - 323) is already wrapped in a check on if the ui is enabled. If there isn't going to be a unique setting to enable the configmap, this is a redundant check.

charts/opencost/templates/configmap-frontend.yaml
```diff
@@ -0,0 +1,93 @@
+{{- if .Values.opencost.ui.enabled }}
```
toscott, line 1: Is this specific configmap always necessary if opencost ui is enabled? My initial reaction is this should be a separate setting, but as long as its not going to be a breaking change for users I'm ok enabling it by default.

toscott (changes requested):
A few comments, but overall looks fine as long as this isn't going to break opencost-ui users. 

Will need to increment the minor version number in the chart to pass the build checks before we can merge.

charts/opencost/templates/configmap-frontend.yaml
mittal-ishaan: This configmap is added during runtime in `opencost-ui` image. Which has write permission issues in openshift environment. This change will not break with opencost-ui `v1.114.0` in which @kastl-ars added an if condition in the entrypoint script [there](https://github.com/opencost/opencost-ui/blob/e1496653bac40db5e5dcce21eb5a6bd55cb25d81/docker-entrypoint.sh#L21-L23). We can get this in after bumping opencost-ui image to 1.114.0 . But as it can break for lower versions, should we keep it under a separate setting, considering that we can release it with ui image bump.

toscott:
The updates look good on this one to me. We will need a version bump here as well. 

I'm thinking we can just coordinate this merge with #245, doing one version bump to 2.0.0 for both of them though.

ameijer (approved):
LGTM

mittal-ishaan:
[Merged](https://github.com/mittal-ishaan/opencost-helm-chart/pull/1) this in https://github.com/opencost/opencost-helm-chart/pull/245 to have a combined major release for `v2.0.0`

PR #247, opened 2025-01-13
Adding value for Linode token

shanduur:
Related to: https://github.com/opencost/opencost/pull/2822
Fixes: https://github.com/opencost/opencost/issues/2128


toscott:
I recommend we hold off on merging any linode configuration until the upstream PR goes in to support it:
https://github.com/opencost/opencost/pull/3037

PR #251, opened 2025-01-24
Welcome Ishaan and Jesse

ameijer:
As voted on in the community meeting on 1/23, we are adding @mittal-ishaan and @jessegoodier as maintainers! Welcome! 

Also, removing Matt Ray who is no longer maintaining the helm chart


toscott (approved):
Sounds good. Happy to have the help

Issue #254, opened 2025-02-03
Add support for API Ingress


josephteddick:
Currently the opencost chart does not support configuration an Ingress for the API service but only the UI:
https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/templates/ingress.yaml

This is useful to users who wish to give external access to the API and it's data.

The fix would entail adding an "apiIngress" to the existing ingress.yaml file with values opencost -> exporter -> apiIngress matching what we have under opencost -> ui -> ingress:
https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/values.yaml#L447

PR #255, opened 2025-02-03
Add support for API Ingress

josephteddick:
This PR intends to allow users to give external access to the API by allowing for configuration of an Ingress.

Fix for issue: https://github.com/opencost/opencost-helm-chart/issues/254

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -373,6 +373,8 @@ opencost:
       namespaceName: prometheus-system
       # -- Service port of in-cluster Prometheus
       port: 80
+      # -- Scheme to use for in-cluster Prometheus
+      scheme: http
     amp:
       # -- Use Amazon Managed Service for Prometheus (AMP)
       enabled: false  # If true, opencost will be configured to remote_write and query from Amazon Managed Service for Prometheus.
@@ -387,6 +389,7 @@ opencost:
         serviceName: my-thanos-query
         namespaceName: opencost
         port: 10901
+        scheme: http
       external:
         enabled: false
         url: 'https://thanos-query.example.com/thanos'
@@ -417,6 +420,8 @@ opencost:
     # used in the default.nginx.conf if you want to switch for using with Docker
     # apiServer: 0.0.0.0
     uiPort: 9090
+    # Set to true to use IPv6
+    useIPv6: true
     # Liveness probe configuration
     livenessProbe:
       # -- Whether probe is enabled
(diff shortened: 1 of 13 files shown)


charts/opencost/templates/ingress.yaml
```diff
@@ -37,4 +37,45 @@ spec:
                   name: {{ $.Values.opencost.ui.ingress.servicePort }}
         {{- end }}
   {{- end }}
+{{- end }}
+---
+{{- if .Values.opencost.exporter.apiIngress.enabled }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "opencost.fullname" . }}-apiIngress
```
TheisFerre, line 46: This name is not valid in kubernetes
```suggestion
  name: {{ include "opencost.fullname" . }}-api-ingress
```

charts/opencost/values.yaml
```diff
@@ -224,6 +224,28 @@ opencost:
...
+      ingressClassName: ""
+      # -- Annotations for Ingress resource
+      annotations: {}
+        # kubernetes.io/tls-acme: "true"
+      # -- A list of host rules used to configure the Ingress
+      # @default -- See [values.yaml](values.yaml)
+      hosts:
+        - host: example.local
+          paths:
+            - /
```
TheisFerre, line 240:
```suggestion
          paths:
            - path: /
              pathType: Prefix
```

charts/opencost/templates/ingress.yaml
```diff
@@ -37,4 +37,45 @@ spec:
...
+    {{- end }}
+  {{- end }}
+  rules:
+    {{- range .Values.opencost.exporter.apiIngress.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .paths }}
+          - path: {{ . }}
+            pathType: Prefix
```
TheisFerre, line 73:
```suggestion
          - path: {{ .path }}
            pathType: {{ .pathType }}
```

TheisFerre (commented):
I support this. Would be a nice addition to the helm chart! I did encounter an issue with the Ingress name though, it has to follow the [kubernetes naming convention](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

I also suggest that we add the option to change the `pathType` as there can be use cases where this will be relevant (i.e. Oauth2 Proxy in front of API)

josephteddick:
@TheisFerre Thanks for reviewing, great suggestions! Just made the changes, let me know if there is anything else.

josephteddick:
@TheisFerre Anything else I need to do to get this merged in?

mittal-ishaan (commented):
thanks for your contribution! can you bump the patch version of the chart?

josephteddick:
@mittal-ishaan Just bumped the patch version. Please let me know if there is anything else you need.

mittal-ishaan (approved):

PR #257, opened 2025-02-11
feat(parquet-exporter): use appVersion in Chart.yaml as default tag for container image

### LINKED ISSUE #259: Use appVersion in Chart.yaml as default tag for container image in parquet exporter
Currently, the parquet exporter chart is using the tag "main" as default for the container image.

The fix, available here : #257 , would rather use the version defined in appVersion of the Chart.yaml.
Besides, the current appVersion does not seem to reference the [latest release](https://github.com/opencost/opencost-parquet-exporter/releases/tag/v0.2.1) of the exporter

peninoule:
This PR aims at using the appVersion defined in Chart.yaml as the default container image rather than using "main". (Same as the opencost helm chart).

It also sets the appVersion to 0.2.1, as it is the [latest release](https://github.com/opencost/opencost-parquet-exporter/releases/tag/v0.2.1) of the exporter

Close #259

### DIFF
--- a/charts/opencost-parquet-exporter/templates/_helpers.tpl
+++ b/charts/opencost-parquet-exporter/templates/_helpers.tpl
@@ -31,6 +31,17 @@ Create chart name and version as used by the chart label.
 {{- end }}
 
 {{/*
+Create default image tag taken from the AppVersion.
+*/}}
+{{- define "opencost-parquet-exporter.imageTag" -}}
+{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}
+{{- end }}
+
+{{- define "opencost-parquet-exporter.fullImageName" -}}
+{{ .Values.image.repository -}}:{{ include "opencost-parquet-exporter.imageTag" . }}
+{{- end }}
+
+{{/*
 Common labels
 */}}
 {{- define "opencost-parquet-exporter.labels" -}}
--- a/charts/opencost-parquet-exporter/templates/_helpers.tpl
+++ b/charts/opencost-parquet-exporter/templates/_helpers.tpl
@@ -34,7 +34,7 @@ Create chart name and version as used by the chart label.
 Create default image tag taken from the AppVersion.
 */}}
 {{- define "opencost-parquet-exporter.imageTag" -}}
-{{ .Values.image.tag | default (printf "%s" .Chart.AppVersion) }}
+{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}
 {{- end }}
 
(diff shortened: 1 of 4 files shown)


charts/opencost-parquet-exporter/templates/_helpers.tpl
```diff
@@ -30,6 +30,17 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Create default image tag taken from the AppVersion.
+*/}}
+{{- define "opencost-parquet-exporter.imageTag" -}}
+{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}
```
mittal-ishaan, line 37:
```suggestion
{{ .Values.image.tag | default (printf "%s" .Chart.AppVersion) }}
```

mittal-ishaan:
@peninoule could you test it once before we can get this in?

peninoule:
> @peninoule could you test it once before we can get this in?

@mittal-ishaan, thanks for the follow up on this.
As is, it won't work with [your commit](https://github.com/opencost/opencost-helm-chart/pull/257/commits/8f6c61964d9f3c8861b365110a58a73808ab37a9), since the opencost-parquet-exporter versions do use a "vX.X.X". Like

charts/opencost-parquet-exporter/templates/_helpers.tpl
```diff
@@ -30,6 +30,17 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Create default image tag taken from the AppVersion.
+*/}}
+{{- define "opencost-parquet-exporter.imageTag" -}}
+{{ .Values.image.tag | default (printf "%s" .Chart.AppVersion) }}
```
mittal-ishaan, line 37:
```suggestion
{{ .Values.image.tag | default (printf "v%s" .Chart.AppVersion) }}
```

peninoule:
So I've tested locally, and I do catch the `v2.0.1` tag when setting the image tag to `""` in my values.yaml.
cronjob works like a charm :+1:

mittal-ishaan:
Image tag you are getting is v0.2.1 right?

peninoule:
@mittal-ishaan v.0.2.1, yes, my bad :sweat_smile:
Thank your for the review !

PR #261, opened 2025-03-03
Updating chart badge

TheUnixRoot:
Updating reference of chart publish workflow as current is not showing real status


mittal-ishaan (approved):
Thank you @TheUnixRoot

PR #264, opened 2025-03-10
Draft: add sha256sums of configMaps to trigger a restart of the pod, if the configMap changes

kastl-ars:
First rough draft!

- refactor annotation section to allow setting annotations for configMaps
- add annotation for configmap-custom-pricing.yaml if .Values.opencost.customPricing.createConfigmap and .Values.opencost.customPricing.enabled are true
- add annotation for configmap-metrics-config.yaml if .Values.opencost.metrics.config.enabled is true


kastl-ars:
> I'll see if I can implement it this week.

That was easier than I thought. @mittal-ishaan Should we also add the secret to the list? To make sure changes to the secret lead to a restart of the deployment?

mittal-ishaan:
sure, we should add the secrets too. Would be better

kastl-ars:
> sure, we should add the secrets too. Would be better

Done.

mittal-ishaan (approved):

Issue #265, opened 2025-03-14
More work has to be done regarding using this on Openshift : Security Context Constraints


v0nNemizez:
Security Context and SCC:
Create template for a SCC that allows containers access to modify internal files.

**Log:**
2025-03-14T09:44:42.797880959Z ERR Controller: pullWatchers: failed to save statuses failed to save config statuses to file: open /var/configs/cloud-configurations.json: permission denied

https://docs.redhat.com/en/documentation/openshift_container_platform/3.11/html/cluster_administration/admin-guide-manage-scc

 _OpenCost Version: 1.110.0 (has also tested with 1.114.0, no error messages but the pod freezes and restart)_

v0nNemizez:
Note: I have already done some work on this so i can contribute when ready

mittal-ishaan:
Thank you @v0nNemizez 
sure, please go ahead with creating a PR and I can help it get it through.

For `Bearer token secret`, though I have not yet tried to reproduce it myself could you help me with your values file when using the secret. 
Also for openshift built-in prometheus, not sure if related but sounds similar, we do have https://github.com/opencost/opencost-helm-chart/blob/3f266a637bb5557925271cfe735578e72c9db552/charts/opencost/values.yaml#L338
that gets the required token and uses it when querying prometheus, and you won't need to set the `bearer_token`

v0nNemizez:
i will test that out... :) i created a PR with a SCC... this one i quite broad, so we might need to adjust it a little... :)

v0nNemizez:
I consider this okay now :)

mittal-ishaan:
Thank you so much on your work here @v0nNemizez 
really appreciated 😄

PR #267, opened 2025-03-17
Openshift Security Context Constraints and updated ClusterRole with access to internal prometheus.

v0nNemizez:
Here is a PR regarding settings specific to Openshift. This is still a work in progress while im testing. in relateion to #265 265


v0nNemizez:
this is ready for review... i got it working on our OpenShift test environment... Check it out, and please give me feedback... :)

charts/opencost/values-openshift.yaml
mittal-ishaan, line 1: Let's have the main values file as the only values file. And have `values-openshift.yaml` as an example values file that users can use to understand the bare minimum configurations needed for an openshift env. 
Having a new values file for openshift that is a duplicate of main values file will make it difficult to maintain and we would need to make changes to both the values file every time.

charts/opencost/templates/deployment.yaml
```diff
@@ -166,6 +166,12 @@ spec:
                 secretKeyRef:
                   name: {{ .Values.opencost.prometheus.existingSecretName | default (include "opencost.prometheus.secretname" .) }}
                   key: {{ .Values.opencost.prometheus.bearer_token_key }}
+            {{- else if and .Values.opencost.platforms.openshift.enabled}}
```
mittal-ishaan, line 169: We already have https://github.com/opencost/opencost-helm-chart/blob/3f266a637bb5557925271cfe735578e72c9db552/charts/opencost/templates/deployment.yaml#L163
why do we need this?

mittal-ishaan (changes requested):
Generally not all openshift users would want these configurations by default. I would suggest we have these new manifests gated with another helm config that users can enable or disable keeping the openshift env. enabled.

charts/opencost/templates/_helpers.tpl
```diff
@@ -191,3 +193,8 @@ apiVersion: networking.k8s.io/v1beta1
 {{- .Values.opencost.ui.image.registry -}}/{{- .Values.opencost.ui.image.repository -}}:{{- include "opencostUi.imageTag" . -}}
 {{- end -}}
 {{- end -}}
+
+{{- define "opencost.sccName" -}}
+opencost-scc
```
mittal-ishaan, line 198:
```suggestion
{{include "opencost.fullname" .}}-scc
```

charts/opencost/templates/scc.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if .Values.opencost.platforms.openshift.enabled }}
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: {{include "opencost.fullname" .}}-scc
```
mittal-ishaan, line 5:
```suggestion
  name: {{include "opencost.sccName" .}}
```

charts/opencost/templates/deployment.yaml
v0nNemizez: We really dont... this was to test createing a secret with only DB_BEARER_TOKEN when you enabled the openshift-config... i solved this another way... i will push a commit to remove it

charts/opencost/values-openshift.yaml
v0nNemizez: okay :) We might want to add a enableSCC then correct under platforms.openshift?

mittal-ishaan: Yes that would be great

charts/opencost/templates/scc.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if .Values.opencost.platforms.openshift.enableSCC }}
```
mittal-ishaan, line 1:
```suggestion
{{- if and .Values.opencost.platforms.openshift.enabled .Values.opencost.platforms.openshift.enableSCC }}
```

mittal-ishaan: At other places we can do it similarly

charts/opencost/templates/clusterrolebinding.yaml
```diff
@@ -39,3 +39,19 @@ subjects:
     namespace: {{ include "opencost.namespace" . }}
 ---
 {{- end }}
+
+{{- if .Values.opencost.platforms.openshift.enableSCC}}
```
mittal-ishaan, line 43:
```suggestion
{{- if and .Values.opencost.platforms.openshift.enabled .Values.opencost.platforms.openshift.enableSCC }}
```

charts/opencost/templates/clusterrole.yaml
```diff
@@ -78,4 +78,16 @@ rules:
       - get
       - list
       - watch
+
+{{- if .Values.opencost.platforms.openshift.enabled }}
```
mittal-ishaan, line 82: @v0nNemizez Just wanted to check—what was the reasoning behind adding this?

mittal-ishaan:
My bad for bringing up this now, just thinking: 
Would it be possible to apply the required OpenShift SCC settings directly in the container spec instead of defining a separate SCC resource? Something like:
https://github.com/opencost/opencost-helm-chart/blob/0bff686322f982ad4e0e3c18e992980975d0453e/charts/opencost/templates/deployment.yaml#L105-L107
Are there any limitations or blockers preventing this approach?

v0nNemizez:
For our part we needed a way to give the pod premissions not granted by our existing SCCs, as i understand it the SCC validates the SecurityContext to see if the serviceaccount is allowed to have that Context. If not, the deployment will ether fail or give errors. Here are a article explaining https://kamsjec.medium.com/what-are-security-contexts-and-security-context-constraints-f17239c5d114

v0nNemizez:
anyhow these settings are toggled by the enableSCC bool in the value file, so it can be activated when needed...

charts/opencost/templates/clusterrole.yaml
mittal-ishaan, line 82: if only needed when using in-cluster openshift prometheus lets gate it with another config

v0nNemizez (commented):
Shoud be okay now

charts/opencost/templates/_helpers.tpl
```diff
@@ -118,7 +119,8 @@ Check that either thanos external or internal is defined
     {{- $ns := .Values.opencost.prometheus.thanos.internal.namespaceName }}
     {{- $clusterName := .Values.clusterName }}
     {{- $port := .Values.opencost.prometheus.thanos.internal.port | int }}
-    {{- printf "http://%s.%s.svc.%s:%d" $host $ns $clusterName $port -}}
+    {{- $scheme := .Values.opencost.prometheus.thanos.internal.scheme}}
```
mittal-ishaan, line 122:
```suggestion
    {{- $scheme := .Values.opencost.prometheus.thanos.internal.scheme | default "http"}}
```

charts/opencost/templates/clusterrole.yaml
```diff
@@ -78,4 +78,19 @@ rules:
       - get
       - list
       - watch
+
+{{- if .Values.opencost.platforms.openshift.enablePromAccess }}
```
mittal-ishaan, line 82:
```suggestion
{{- if and .Values.opencost.platforms.openshift.enabled .Values.opencost.platforms.openshift.enablePromAccess }}
```

charts/opencost/templates/_helpers.tpl
```diff
@@ -102,8 +102,9 @@ Create the name of the controller service account to use
     {{- $host := tpl .Values.opencost.prometheus.internal.serviceName . }}
     {{- $ns := tpl .Values.opencost.prometheus.internal.namespaceName . }}
     {{- $clusterName := .Values.clusterName }}
+    {{- $scheme := .Values.opencost.prometheus.internal.scheme | default "http"}}
```
mittal-ishaan, line 105: we can add this and thanos scheme config in our helm values file too.

mittal-ishaan: prometheus internal is missing the scheme config. Please can you add that.

mittal-ishaan (approved):
This looks great now. @v0nNemizez do check my last comment and we can merge it. Also could you add some description of how you tested it.

mittal-ishaan:
@v0nNemizez please can you add the `scheme` in `opencost.promtheus.internal` too. otherwise the PR is good to merge

PR #271, opened 2025-03-31
allow: set path for internal prometheus

dodizzle:
When querying prometheus  behind a reverse proxy such as [The mimir-query-frontend](https://grafana.com/docs/mimir/latest/references/architecture/components/query-frontend/) we need to specify the path to prometheus.
ie /prometheus
If prometheus is external then you can set it in the url:
opencost.prometheus.external.url: https://example.com/prometheus
But if you are querying prometheus in the cluster there currently is no option.
This PR resolves that issue by allowing it to be set here -> opencost.prometheus.internal.path

### DIFF
--- a/charts/opencost/templates/_helpers.tpl
+++ b/charts/opencost/templates/_helpers.tpl
@@ -98,13 +98,18 @@ Create the name of the controller service account to use
     {{- $port := .Values.opencost.sigV4Proxy.port | int }}
     {{- $ws := .Values.opencost.prometheus.amp.workspaceId }}
     {{- printf "http://localhost:%d/workspaces/%v" $port $ws -}}
-  {{- else -}}
+{{- else -}}
     {{- $host := tpl .Values.opencost.prometheus.internal.serviceName . }}
     {{- $ns := tpl .Values.opencost.prometheus.internal.namespaceName . }}
     {{- $clusterName := .Values.clusterName }}
     {{- $port := .Values.opencost.prometheus.internal.port | int }}
-    {{- printf "http://%s.%s.svc.%s:%d" $host $ns $clusterName $port -}}
-  {{- end -}}
+    {{- $path := .Values.opencost.prometheus.internal.path | default "" }}
+    {{- if $path }}
+      {{- printf "http://%s.%s.svc.%s:%d%s" $host $ns $clusterName $port $path -}}
+    {{- else }}
+      {{- printf "http://%s.%s.svc.%s:%d" $host $ns $clusterName $port -}}
+    {{- end }}
+{{- end -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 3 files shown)


ameijer (commented):
thanks for your contribution! can you bump the patch version of the chart?

mittal-ishaan:
@dodizzle can you resolve the merge conflicts and bump the chart's patch version? The changes overall look good

charts/opencost/templates/_helpers.tpl
```diff
@@ -98,14 +98,19 @@ Create the name of the controller service account to use
...
     {{- $host := tpl .Values.opencost.prometheus.internal.serviceName . }}
     {{- $ns := tpl .Values.opencost.prometheus.internal.namespaceName . }}
     {{- $clusterName := .Values.clusterName }}
     {{- $scheme := .Values.opencost.prometheus.internal.scheme | default "http"}}
     {{- $port := .Values.opencost.prometheus.internal.port | int }}
-    {{- printf "%s://%s.%s.svc.%s:%d" $scheme $host $ns $clusterName $port -}}
-  {{- end -}}
+    {{- $path := .Values.opencost.prometheus.internal.path | default "" }}
+    {{- if $path }}
+      {{- printf "http://%s.%s.svc.%s:%d%s" $host $ns $clusterName $port $path -}}
```
mittal-ishaan, line 109:
```suggestion
      {{- printf "%s://%s.%s.svc.%s:%d%s" $scheme $host $ns $clusterName $port $path -}}
```

charts/opencost/templates/_helpers.tpl
```diff
@@ -98,14 +98,19 @@ Create the name of the controller service account to use
...
+    {{- else }}
+      {{- printf "http://%s.%s.svc.%s:%d" $ scheme $host $ns $clusterName $port -}}
```
mittal-ishaan, line 111:
```suggestion
      {{- printf "%s://%s.%s.svc.%s:%d" $scheme $host $ns $clusterName $port -}}
```

mittal-ishaan (commented):
Minor merge conflict changes. Can you test it once after you correct them now?

mittal-ishaan:
@dodizzle Can you bump the chart's patch version and test it once to confirm before we can merge it?

ameijer:
@dodizzle would love to get this merged!

edwardccg:
My current workaround for this are following after looking the helper code:

```yaml
prometheus:
  # we use the external url but its pointing to internal url as we have custom /prometheus path, the internal version doesn't support custom path.
  # At the end of day, these config are just setting up PROMETHEUS_SERVER_ENDPOINT env variable
  external:
    enabled: true
    url: http://promstack-kube-prometheus-prometheus.observability.svc.cluster.local:9090/prometheus
  internal:
    enabled: false
```

PR #272, opened 2025-04-16
Add opencost.ui.useIPv6 feature flag

tintii:
Currently errors are thrown when the nginx is unable to bind to an IPv6 address. This needs to be disabled for clusters where IPv6 is disabled.

I've added a flag that defaults to true for this setting so it's backwards compatibly but allows disabling the IPv6 requirement.

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 2.0.1
+version: 2.1.0
 maintainers:
   - name: jessegoodier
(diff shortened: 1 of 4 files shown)


charts/opencost/Chart.yaml
```diff
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 2.0.1
+version: 2.1.0
```
ameijer, line 12: can you bump the patch version instead? thanks for your contribution!

mittal-ishaan (approved):

PR #274, opened 2025-04-21
Fix for https://github.com/opencost/opencost-helm-chart/pull/272 to make feature flag actually work.

tintii:
My apologies, turns out the extra defaults withing the configmap was.. well.. extra.

This fixes it and the feature flag works now.

Cheers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+.idea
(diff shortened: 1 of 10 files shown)


mittal-ishaan (approved):
Thank you so much for checking on this.

PR #275, opened 2025-05-29
Promless Config

Sean-Holcomb:
This PR adds configuration which allow for the use of the Collector Datasource (promless). Additionally it adds in configuration for persistent volume path which the Collector Datasource can utilize to preserve states between restarts.

This was tested by templating the different values and deploying Opencost in various permutations with the Collector Datasource enabled.

### DIFF
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -115,26 +115,26 @@ spec:
             {{- with .Values.opencost.exporter.collectorDataSource }}
             {{- if .enabled }}
             - name: COLLECTOR_DATA_SOURCE_ENABLED
-              value: true
+              value: "true"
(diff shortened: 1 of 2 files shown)


ameijer (approved):

charts/opencost/templates/deployment.yaml
```diff
@@ -112,6 +112,32 @@ spec:
               value: {{ .Values.plugins.enabled | quote }}
             - name: KUBECOST_NAMESPACE
               value: {{ include "opencost.namespace" . }}
+            {{- with .Values.opencost.exporter.collectorDataSource }}
```
mittal-ishaan, line 115: we need these configs in the values file.

Sean-Holcomb: I have added the values and some notes about them

PR #282, opened 2025-06-09
Add ability to configure resolution for prometheus

Sean-Holcomb:
Add environment variable which allows configuration of query resolution
https://github.com/opencost/opencost/pull/3182


Sean-Holcomb:
Yeah, if you don't mind giving it a review

ameijer (approved):
LGTM, cc @mittal-ishaan to merge

mittal-ishaan:
Yes, as discussed, considering that top projects like grafana and bitnami follow the release process as we have right now, we are sticking to it and we would require a version bump here.

PR #283, opened 2025-06-23
feat(sec) - customize service account mounting

cpsmx:
It's best practice to disable automounting the service account tokens. In heavily audited environments there is often an admission controller preventing pods starting that have automountServiceAccountToken enabled and the same with service accounts. They require it to be explicitly set to false.

This change makes that behaviour configurable per SA and pod.


ameijer (approved):
LGTM

PR #284, opened 2025-06-28
(doc) update readme to easily install unittest

karthik-suresh:
Looks a good practice to install plugins and an easier one-liner to install unittest instead of visiting the unittest URL for more documentation.


mittal-ishaan:
Thank you for the fix

PR #285, opened 2025-07-15
Change ETL env variable name

Sean-Holcomb:
Change name Prometheus query resolution env variable
https://github.com/opencost/opencost/pull/3234


ameijer (approved):

mittal-ishaan:
Hi @Sean-Holcomb,
Can you bump the chart's patch version? and we can go ahead with merging it.

Issue #286, opened 2025-07-21
Add support for Pod Disruption Budget


josephteddick:
While the OpenCost helm chart does have support for affinity and multiple replicas, it does not for Pod Disruption Budgets. This final piece will allow customization in making OpenCost highly available.

I propose adding a `pdb` section to the manifest with options to customize (minAvailable and maxUnavailable).
https://github.com/opencost/opencost-helm-chart/blob/main/charts/opencost/values.yaml#L118

```yaml
pdb:
  enabled: false
  # -- Minimum number of pods that must be available after the eviction
  minAvailable: ~
  # -- Maximum number of pods that can be unavailable after the eviction
  maxUnavailable: ~
```

My personal use case will be enabling and setting minAvailable to 1, to ensure there is always one pod running with matching selector labels (in conjunction with affinity and multiple replicas).

PR #287, opened 2025-07-21
Add support for Pod Disruption Budget

josephteddick:
This PR intends to allow users to make OpenCost highly available by configuring a Pod Disruption Budget (affinity and replicas already present).

Fix for issue: https://github.com/opencost/opencost-helm-chart/issues/286

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 2.1.6
+version: 2.1.7
 maintainers:
   - name: jessegoodier
   - name: toscott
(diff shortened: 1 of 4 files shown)


mittal-ishaan (approved):
Thank you @josephteddick

ameijer:
@josephteddick , does this mean when you run multiple replicas, you have multiple prom scrape targets, one for each OC? PR LGTM, just trying to understand

josephteddick:
> @josephteddick , does this mean when you run multiple replicas, you have multiple prom scrape targets, one for each OC? PR LGTM, just trying to understand

Nope, each replicas has the same setup in that they are pointing to the service name for Prometheus. It's then up to internal routing for which pods (replica) it will go to and fetch the metrics (assuming Prometheus is also HA). You also don't have to worry about duplication because each replica will have it's own PVs.

High Availability here is mostly for the OpenCost API in that we can always hit it to get allocation metrics. Hope that helps.

PR #290, opened 2025-07-29
add option to override the default container command

nishanthreddydd:
We can run opencost in two modes, in agent mode only metrics are emitted and in the default mode it provides all the cost attribution api's in addition to cost metrics. But the helm chart doesn't provide an option to run opencost in `agent` mode. This PR adds that option.


nishanthreddydd:
@mittal-ishaan bump'd the patch version.

PR #292, opened 2025-08-08
OpenShift SCC RoleBinding

jlopes-randoli:
I found a minor issue that affects the functionality of Opencost while testing on OpenShift.

When `platforms.openshift.enableSCC ` is set to true, we create an SCC and a RoleBinding which resolves permission issues that occur that lead to problems such as not being able to properly fetch pricing data.

However, when I tested, I still encountered these issues despite having this flag enabled. I realized that the Role that the RoleBinding is attempting to reference is actually a ClusterRole. I believe all SecurityContextConstraints that get created, have a corresponding ClusterRole, not Role, associated with it.

When I switched the resource type, the flag ran as expected and my issues were resolved.

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
 - finops
 - monitoring
 - opencost
-version: 2.3.0
+version: 2.3.1
 maintainers:
 - name: jessegoodier
 - name: toscott
--- a/charts/opencost/templates/deployment.yaml
+++ b/charts/opencost/templates/deployment.yaml
@@ -198,8 +198,17 @@ spec:
                   name: {{ include "opencost.prometheus.secretname" . }}
                   key: CLOUD_PROVIDER_API_KEY
             {{- end }}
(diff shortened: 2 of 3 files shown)


mittal-ishaan:
Hi @jlopes-randoli 
we would need a chart patch version update.

I am not sure if SecurityContextConstraints when created have a corresponding cluster role.
The [docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.10/html/authentication_and_authorization/managing-pod-security-policies#role-based-access-to-ssc_configuring-internal-oauth) say we need to create a role that can use the SCC and then bind it to the service account.
Found this [medium](https://medium.com/@tamber/rbac-scc-integration-using-rolebinding-clusterrolebinding-536c069cc6ef) blog too around it.

jlopes-randoli:
@mittal-ishaan That's a good catch, I tested again and I believe there was a prior command I ran that might've created that ClusterRole but it's not created automatically.

I'm going to do some more testing and update the PR with the actual Role as well once I've confirmed.

PR #297, opened 2025-08-21
Update env var names and values

Sean-Holcomb:
This PR is part of an effort to standardize retention value and env vars across projects. I have added backwards compatibility for existing values


mittal-ishaan (approved):
@Sean-Holcomb just needs a patch version bump

PR #301, opened 2025-09-04
Dodizzle/proxy fix

### LINKED ISSUE #270: Add path as an option for prometheus.internal if behind a reverse proxy(mimir)
When querying prometheus  behind a reverse proxy such as [The mimir-query-frontend](https://grafana.com/docs/mimir/latest/references/architecture/components/query-frontend/) we need to specify the path to prometheus.
ie /prometheus
If prometheus is external then you can set it in the url:
opencost.prometheus.external.url: https://example.com/prometheus
But if you are querying prometheus in the cluster there currently is no option.
This PR resolves that issue by allowing it to be set here -> opencost.prometheus.internal.path

ameijer:
pushes PR #271 over the line. Closes #270


mittal-ishaan (approved):
Thank you @ameijer 
jist a patch version bump is needed

PR #307, opened 2025-09-30
feat: Add option to use cm to set CLUSTER_ID envvar

gracedo:
Instead of a predefined default cluster id value, there should be an option to set this dynamically. I've added the option here to use a user-generated configmap to set the CLUSTER_ID envvar. User is responsible for generating the configmap prior to installation.

Ported over from https://github.com/kubecost/kubecost/pull/368


ameijer (approved):
LGTM, thanks for the contribution!

PR #310, opened 2025-10-04
OpenShift SCC RoleBinding

jlopes-randoli:
Re-opening this PR (https://github.com/opencost/opencost-helm-chart/pull/292)

Fixed my solution to also create the ClusterRole that grants permission to use the SCC that gets created. Tested this on an OpenShift cluster and is working as intended.

@mittal-ishaan Can you explain the chart patch version update you mentioned on my previous PR?


ameijer:
closing, ishaan will create a new PR

PR #313, opened 2025-10-17
initial opencost Helm values schema

HMetcalfeW:
Interested in the team's feedback on adding a schema to our Helm Chart. The initial generation was relatively straightforward, but as our chart evolves, this file will require ongoing maintenance. If we think this is too much maintenance, then I'm happy to table it, but believe it can pay long-term dividends. This first iteration is lax because it lacks required fields.

The practice is fairly common in the [Bitnami Charts community](https://github.com/search?q=repo%3Abitnami%2Fcharts+%22values.schema.json%22+language%3AJSON&type=code&l=JSON). There's a great blog post on the motivations behind creating a [Values schema](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) and the various checks it allows.

Requirement checks. Example: An API_KEY environment variable is set
Type validation. Example: The image tag is a string such as "1.5" and not the number 1.5
Range validation. Example: The value for a CPU utilization percentage key is between 1 and 100
Constraint Validation. Example: The pullPolicy is IfNotPresent, Always, or Never; A URL has the format http(s)://:
Since schema checks are run on all Helm commands (template, lint, upgrade, install), this helps prevent customer values from drifting too far between versions and clearly defines what values are/are not allowed.

Example Helm output using a number for clusterId
`helm lint .`

```
[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
opencost:
- at '/nameOverride': got number, want string
```

https://helm.sh/docs/helm/helm_lint/

### DIFF
--- a/.github/configs/ct.yaml
+++ b/.github/configs/ct.yaml
@@ -6,7 +6,7 @@ target-branch: main
 chart-dirs:
   - charts
 chart-repos: []
-validate-chart-schema: false
+validate-chart-schema: true
 validate-maintainers: true
 validate-yaml: true
 exclude-deprecated: true
--- a/.github/workflows/helm-test.yml
+++ b/.github/workflows/helm-test.yml
@@ -3,11 +3,11 @@ name: CI
 
 # Controls when the workflow will run
 on:
-  # Triggers the workflow on push or pull request events but only for the "main" branch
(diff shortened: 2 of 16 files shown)


ameijer:
might want to enable strict schema, ensure we validate in CI with CI vals

PR #319, opened 2025-11-13
fix: mcp disable procedure

marijus-ravickas:
If MCP is `enabled: false` then ENV variable `MCP_SERVER_ENABLED` needs to be set to `false`. In other case helm will just remove k8s service, but MCP still start in the container. This PR fixes this.


ameijer:
@marijus-ravickas can you bump patch version of chart?

PR #344, opened 2026-04-17
feat: Add plugins.install.plugins list and existingSecret support (adopts #328)

ameijer:
## Summary

Adopts and completes [opencost/opencost-helm-chart#328](https://github.com/opencost/opencost-helm-chart/pull/328) (originally opened by @ermeaney, now abandoned). Addresses the outstanding review feedback from @ameijer and fixes a regression introduced in the original patch.

It decouples plugin binary downloads from plugin credentials, enabling secure secret management workflows (ExternalSecrets, Vault, etc.).

- Add `plugins.install.plugins` list to specify which plugins to download independently of configs
- Add `plugins.existingSecret` to use an existing secret instead of creating one from values
- Maintains full backward compatibility — existing configs-based workflows continue to work unchanged

## Problem

The current plugin installation mechanism requires plugin configurations to be defined in `plugins.configs` for the install script to download the plugin binary. This creates a security issue because:

1. The install script iterates over `plugins.configs` to determine which plugins to download.
2. Plugin configs contain secrets (e.g., Datadog API key, App key).
3. Values files are often committed to git, exposing credentials.

This makes it impossible to use Kubernetes ExternalSecrets, Vault, or other secret management tools for plugin credentials.

### New Values

```yaml
plugins:
  enabled: true
  install:
    plugins:        # NEW: list of plugins to download (independent of configs)
      - datadog
  existingSecret: "my-datadog-plugin-secret"  # NEW: use an existing secret
  # configs not needed - credentials come from existingSecret
```

### Behavior

- If `plugins.install.plugins` is set, download those plugins regardless of what's in configs.
- If `plugins.install.plugins` is empty, fall back to downloading plugins based on keys in configs (legacy behavior).
- If `plugins.existingSecret` is set, mount that secret instead of creating one from configs.
- If `plugins.existingSecret` is not set, create the secret from configs (legacy behavior).

## Changes from the original PR (#328)

1. **Addressed @ameijer's review**: replaced the `default` call in `deployment.yaml` with dedicated template helpers. Introduced two helpers in `_helpers.tpl`:
   - `opencost.pluginsConfigSecretName` — resolves the plugin secret name (existingSecret or generated).
   - `opencost.plugins.installList` — resolves the list of plugins to install/mount (explicit `install.plugins` list, then configs keys).
2. **Fixed a regression**: the earlier revision guarded `plugins-config.yaml` with `.Values.plugins.configs`, which caused the deployment volume to reference a missing secret when `plugins.enabled=true` with no configs (a valid case in upstream main). Dropped the `configs` condition; the secret is now always rendered when plugins are enabled and no `existingSecret` is set, matching prior behavior.
3. **Added test coverage** for the plugin install list, existingSecret wiring, and the regression scenario above (now 13 helm-unittest cases in `tests/plugins_test.yaml`).
4. **Updated `charts/opencost/README.md`** entries for the two new values so `helm-docs` stays in sync.
5. **Wired `helm unittest` into CI** (`.github/workflows/helm-test.yml`) so the plugin tests (and future unit tests) actually run on every PR. Previously the `tests/*_test.yaml` suites were not executed by CI — only `ct lint` / `ct install` ran.
6. **Bumped chart version** `2.5.14` → `2.5.15`.
7. **Addressed all Copilot review feedback** — see "Responses to Copilot review" below.

## Responses to Copilot review

All four comments accepted. The review-comment reply API isn't writable from this automation, so the responses are recorded here.

### 1. `charts/opencost/README.md` — call out required config keys in the plugins install list doc

> The README entry for `plugins.install.plugins` describes download behavior but not that this list also determines which config filenames the Deployment will mount...

**Accepted.** The README is generated from the `-- ` comments in `values.yaml` via `helm-docs`, so I updated the source comment to spell out the config-source requirement and the Pod-start failure mode, then resynced the `README.md` row. The comment now makes clear that each listed plugin must have a matching config source (either `plugins.configs.<plugin>` or a `<plugin>_config.json` key in `plugins.existingSecret`), and notes that the chart now validates this at template-render time when `existingSecret` is empty.

### 2. `charts/opencost/templates/plugins-config.yaml` — fail on mutually-exclusive `existingSecret` + `configs`

> `plugins.existingSecret` and `plugins.configs` are effectively mutually exclusive... Consider adding an explicit validation (e.g., `fail`) when both are set, similar to the existing mutual-exclusion check for cloud integration values...

**Accepted.** Added a new `opencost.plugins.configCheck` helper in `_helpers.tpl` modeled on `opencost.cloudIntegration.secretConfigCheck`. When both values are set it produces:

> `plugins.existingSecret and plugins.configs are mutually exclusive. When plugins.existingSecret is set the chart does not render a generated Secret from plugins.configs, so any configs entries would be silently ignored. Please specify only one.`

The check is invoked from both `deployment.yaml` and `plugins-config.yaml` so it fires regardless of which template is first rendered, and a new helm-unittest case (`should fail when plugins.existingSecret and plugins.configs are both set`) asserts the message.

### 3. `charts/opencost/values.yaml` — document required config keys for `plugins.install.plugins`

> The new `plugins.install.plugins` value is documented here as controlling downloads only, but it also drives which `<plugin>_config.json` files the Deployment attempts to mount... Please update this value's documentation to explicitly mention the required config keys/files and the startup failure mode (or adjust the templates so mounting is based on available configs rather than the install list).

**Accepted.** Expanded the `-- ` comment in `values.yaml` for both `plugins.install.plugins` and `plugins.existingSecret` to document the `<plugin>_config.json` requirement and the Pod-start failure mode, and noted that the new `configCheck` helper now validates this at template-render time (when `existingSecret` is empty). I deliberately kept the alternative of "mount by available configs instead of the install list" off the table — the whole point of `plugins.install.plugins` is to decouple the install/mount set from `configs`, and silently dropping entries not present in `configs` would defeat the feature.

### 4. `charts/opencost/templates/_helpers.tpl` — validate that `install.plugins` ⊆ `plugins.configs` keys

> `opencost.plugins.installList` can return plugin names that don't exist in `.Values.plugins.configs`... Consider adding a template-time validation that, when `plugins.existingSecret` is empty, `plugins.install.plugins` (if set) is a subset of the keys in `plugins.configs`, and `fail` with an actionable error if not.

**Accepted.** Rolled this into the same `opencost.plugins.configCheck` helper. When `plugins.existingSecret` is empty and `plugins.install.plugins` references a plugin absent from `plugins.configs`, rendering fails with (for example):

> `plugins.install.plugins contains "mongodb" but plugins.configs has no matching "mongodb" entry. The Deployment mounts mongodb_config.json from the generated plugin secret, so the Pod will fail to start with a missing subPath. Either add a plugins.configs.mongodb entry or set plugins.existingSecret to a Secret that contains mongodb_config.json.`

I intentionally scoped the subset check to the `existingSecret == ""` case — the chart cannot introspect an externally-managed Secret at template-render time, so enforcing the check there would produce false positives for perfectly valid configurations. Added two helm-unittest cases: one that asserts the `fail` for the misconfiguration, and one that asserts rendering still succeeds when `existingSecret` is set (bypass path).

## Test plan

- [x] `helm lint charts/opencost` passes
- [x] `helm unittest charts/opencost -f 'tests/plugins_test.yaml'` passes (13/13)
- [x] `helm template` verified for new style (`existingSecret` + `install.plugins`)
- [x] `helm template` verified for old style (`configs` only) — backward compatible
- [x] `helm template` verified for the `plugins.enabled=true` with no configs edge case
- [x] `helm template` verified that the two new `configCheck` failure paths trigger with actionable messages
- [x] CI now runs the helm-unittest plugin suite on every PR

Fixes #327
Supersedes #328

---

Credit to @ermeaney for the original implementation in #328.

<div><a href="https://cursor.com/agents/bc-170c67af-5d27-41bd-b830-6df45848b81e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-170c67af-5d27-41bd-b830-6df45848b81e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


mittal-ishaan (approved):
This looks good. Thanks a lot @ameijer 
small call out: the DCO test is failing, its not required though so we are good to merge.

PR #345, opened 2026-04-20
feat: add OCI cloud cost configuration example to cloudIntegrationJSON

Kush172005:
## Summary

This PR adds an OCI configuration example to the `cloudIntegrationJSON` section in `values.yaml`.

OCI Cloud Costs support already exists in OpenCost (see #3003), but the lack of documentation/examples makes it difficult for users to discover and configure.

## Changes
- Added OCI configuration example under `cloudIntegrationJSON`
- Matches structure of existing providers (AWS, Azure, GCP)

## Why this matters
Users rely on this section as a reference when configuring cloud integrations. Including OCI improves discoverability and usability.

## Testing
- Verified YAML structure
- No breaking changes

Closes #3003

### DIFF
--- a/charts/opencost/values.yaml
+++ b/charts/opencost/values.yaml
@@ -200,7 +200,23 @@ opencost:
   #           "client_x509_cert_url": "my-x509-cert-url"
   #         }
   #       }
-  #     ]
+  #     ],
+  #     "oci": {
+  #       "usageApi": [
+  #         {
+  #           "tenancyID": "ocid1.tenancy.oc1..my-tenancy-ocid",
+  #           "region": "us-ashburn-1",
+  #           "authorizer": {
+  #             "authorizerType": "OCIRawConfigProvider",
+  #             "tenancyId": "ocid1.tenancy.oc1..my-tenancy-ocid",
+  #             "userId": "ocid1.user.oc1..my-user-ocid",
+  #             "region": "us-ashburn-1",
+  #             "fingerprint": "xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx",
+  #             "privateKey": "-----BEGIN RSA PRIVATE KEY-----\nmy-pem-encoded-private-key\n-----END RSA PRIVATE KEY-----"
+  #           }
+  #         }
+  #       ]
+  #     }
   #   }
 
   # -- MCP (Model Context Protocol) Server Configuration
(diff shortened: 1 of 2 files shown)


charts/opencost/values.yaml
```diff
@@ -200,7 +203,23 @@ opencost:
...
+  #         {
+  #           "tenancyID": "ocid1.tenancy.oc1..my-tenancy-ocid",
+  #           "region": "us-ashburn-1",
+  #           "authorizer": {
+  #             "authorizerType": "OCIRawConfigProvider",
+  #             "tenancyId": "ocid1.tenancy.oc1..my-tenancy-ocid",
+  #             "userId": "ocid1.user.oc1..my-user-ocid",
+  #             "region": "us-ashburn-1",
+  #             "fingerprint": "xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx",
+  #             "privateKey": "-----BEGIN RSA PRIVATE KEY-----\nmy-pem-encoded-private-key\n-----END RSA PRIVATE KEY-----"
```
ameijer: can you add a different placeholder here? the reason is that a lot of security scanning tools search for the string "PRIVATE KEY"

Kush172005: Hi @ameijer, I’ve updated the placeholder to avoid triggering security scanners. Thanks for pointing!

charts/opencost/values.yaml
```diff
@@ -162,6 +162,9 @@ opencost:
   # -- "<fullname>-cloud-integration" in the release namespace. Mutually exclusive with
   # -- opencost.cloudIntegrationSecret.
   cloudIntegrationJSON: ""
+  # For OCI, cloud-integration.json uses "oci" as an object with a "usageApi" array (not a
```
mittal-ishaan: @Kush172005 can you remove this comment, we would want this to be part of the docs instead  as it can get messy in here.

mittal-ishaan:
@Kush172005 Can you also bump the patch version of the helm chart

charts/opencost/values.yaml
Kush172005: Thanks @mittal-ishaan i had removed this!

Kush172005:
Removed the comment and bumped the patch version @mittal-ishaan . Let me know if anything else is needed. Thanks!

mittal-ishaan (approved):
Thank you @Kush172005

Issue #367, opened 2026-06-28
Add configurable timeoutSeconds and successThreshold for exporter probes


aman-kumar29:
Summary

The OpenCost Helm chart already exposes startup, liveness, and readiness probe settings for the exporter, but it does not allow users to configure common Kubernetes probe fields like `timeoutSeconds` and `successThreshold`.

This makes it harder to tune probes for slower startup times, heavier workloads, or environments with stricter timing requirements.

Proposal

Add support for the following values under the existing exporter probe configuration:

- `opencost.exporter.startupProbe.timeoutSeconds`
- `opencost.exporter.livenessProbe.timeoutSeconds`
- `opencost.exporter.readinessProbe.timeoutSeconds`
- `opencost.exporter.readinessProbe.successThreshold`

These values should be rendered into the generated Kubernetes probe specs when set.

Why this is useful

- Helps users tune probes for slower or larger deployments
- Aligns the chart with standard Kubernetes probe configuration options
- Keeps the change small and backwards-compatible

Suggested implementation scope

- Update values.yaml to add these fields with comments
- Update deployment templates to render them into the probe definitions
- Add chart tests to cover the new rendering behavior

PR #369, opened 2026-06-28
add timeout configuration for override in probes

aman-kumar29:
These values are just explicit chart defaults. They preserve the current behavior because the defaults match the existing effective probe settings, while still allowing users to override them if needed.
issue - 

issue #367


mittal-ishaan (approved):
Thank you @aman-kumar29

PR #378, opened 2026-08-07
feat: add opencost.exporter.extraEnvFrom to source env from ConfigMap/Secret

ahauserv:
`opencost.exporter.extraEnv` only renders literal name/value pairs, so an exporter env var whose value is not known when the values file is written cannot be supplied.

Typical case: values that are only known in the cluster at deploy time, or values that belong in a Secret rather than in chart values.

```yaml
opencost:
  exporter:
    extraEnvFrom:
      - configMapRef:
          name: opencost-env
      - secretRef:
          name: opencost-env-secret
```
Explicit `env:` entries win over `envFrom` in Kubernetes, so everything the chart sets explicitly (including `extraEnv`) keeps precedence.

Chart version bumped to 2.5.30; README table row and version badge updated accordingly.

## How was this tested

`helm template` with the values above renders on the exporter container:

```yaml
envFrom:
  - configMapRef:
      name: opencost-env
  - secretRef:
      name: opencost-env-secret
```

With the value unset, no `envFrom` is rendered (unchanged output).
`helm lint` and `helm unittest charts/opencost -f 'tests/plugins_test.yaml'` pass.

### DIFF
--- a/charts/opencost/Chart.yaml
+++ b/charts/opencost/Chart.yaml
@@ -9,7 +9,7 @@ keywords:
   - finops
   - monitoring
   - opencost
-version: 2.5.29
+version: 2.5.30
 maintainers:
   - name: jessegoodier
   - name: toscott
--- a/charts/opencost/README.md
+++ b/charts/opencost/README.md
@@ -2,7 +2,7 @@
 
 OpenCost and OpenCost UI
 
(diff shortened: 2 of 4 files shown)


ahauserv:
@ameijer a small change that would make it possible to set ENVs from a secret/configmap. The workflow is waiting for an approval.

thomasvn (approved):
Looks good, thank you @ahauserv!

